PR #22, opened 2022-03-17
Fix leak in varargs

cpwright:


src/main/c/jpy_jobj.c
```diff
@@ -145,18 +146,20 @@ int JObj_init(JPy_JObj* self, PyObject* args, PyObject* kwds)
...
 
-    objectRef = (*jenv)->NewGlobalRef(jenv, objectRef);
-    if (objectRef == NULL) {
+    globalObjectRef = (*jenv)->NewGlobalRef(jenv, localObjectRef);
+    if (globalObjectRef == NULL) {
         PyErr_NoMemory();
         return -1;
     }
+    (*jenv)->DeleteLocalRef(jenv, localObjectRef);
+    localObjectRef = NULL;
```
devinrsmith, line 155: Been a while since I've dug into JNI; are we being extra careful here by deleting the local ref? In which case, do we also want to delete it in the if block?

On the other hand, I think we may be allowed to be sloppier and not clean up local refs: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#local_references

> Local references are valid for the duration of a native method call. They are freed automatically after the native method returns.

src/main/c/jpy_jtype.c
```diff
@@ -269,7 +271,21 @@ JPy_JType* JType_New(JNIEnv* jenv, jclass classRef, jboolean resolve)
...
+        return NULL;
+    }
     type->isInterface = (*jenv)->CallBooleanMethod(jenv, type->classRef, JPy_Class_IsInterface_MID);
+    if ((*jenv)->ExceptionCheck(jenv)) {
+        (*jenv)->ExceptionClear(jenv);
+        PyMem_Del(type->javaName);
+        type->javaName = NULL;
+        metaType->tp_free(type);
+        return NULL;
+    }
```
devinrsmith, line 288: I'm curious when/why we'd expect exceptions with these method calls.

src/main/c/jpy_jtype.c
```diff
@@ -1103,8 +1131,11 @@ int JType_ProcessClassMethods(JNIEnv* jenv, JPy_JType* type)
         // we exclude bridge methods; as covariant return types will result in bridge methods that cause ambiguity
         if (isPublic && !isBridge) {
             methodNameStr = (*jenv)->CallObjectMethod(jenv, method, JPy_Method_GetName_MID);
+            JPy_ON_JAVA_EXCEPTION_RETURN(-1);
             returnType = (*jenv)->CallObjectMethod(jenv, method, JPy_Method_GetReturnType_MID);
+            JPy_ON_JAVA_EXCEPTION_RETURN(-1);
             parameterTypes = (*jenv)->CallObjectMethod(jenv, method, JPy_Method_GetParameterTypes_MID);
+            JPy_ON_JAVA_EXCEPTION_RETURN(-1);
             mid = (*jenv)->FromReflectedMethod(jenv, method);
```
devinrsmith, line 1139: Are these new error cases that you've seen? Or, just trying to follow best-practices?

src/main/c/jpy_jtype.c
```diff
@@ -1599,6 +1635,7 @@ int JType_MatchVarArgPyArgAsJBooleanParam(JNIEnv *jenv, JPy_ParamDescriptor *par
         else return 0;
         minMatch = matchValue < minMatch ? matchValue : minMatch;
     }
+    Py_XDECREF(varArgs);
```
devinrsmith, line 1638: It looks like there are cases where we return early (in other instances as well) and won't decref.

src/main/c/jpy_jobj.c
cpwright: You can let the local refs leak, but it is better to not do that if you can avoid it.  When you run with -XCheck:jni it will tell you when you exceed the allocated number of local refs; which we were doing quite often.  This resulted in having hundreds of local refs (when the default local ref stack is 16 or 32).

src/main/c/jpy_jtype.c
cpwright: I generally would not expect an exception, but this is cleanup that XCheck:jni indicated we were missing the exception check, so if we did have an exception it would show up randomly later.

src/main/c/jpy_jtype.c
cpwright: Best practices.

src/main/c/jpy_jtype.c
```diff
@@ -1837,6 +1876,8 @@ int JType_ConvertVarArgPyArgToJObjectArg(JNIEnv* jenv, JPy_ParamDescriptor* para
         }
     }
 
+    Py_XDECREF(pyArg);
+
```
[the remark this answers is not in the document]
cpwright: I will review these again, because the error handling path here will really matter in terms of leaks.  For the local refs leaked from the JNI I didn't care too much on the error path, but these python references matter (as would JNI GlobalRefs).

src/main/c/jpy_jtype.c
```diff
@@ -1359,6 +1390,7 @@ JPy_ParamDescriptor* JType_CreateParamDescriptors(JNIEnv* jenv, int paramCount,
         paramDescriptor = paramDescriptors + i;
 
         type = JType_GetType(jenv, paramClass, JNI_FALSE);
+        (*jenv)->DeleteLocalRef(jenv, paramClass);
```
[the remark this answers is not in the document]
cpwright: I could do it with a MACRO, but I am nervous about macros that behave sort of like functions, but not in that you would evaluate the reference twice and there can be danger.  Since it is being used as an LHS, it is probably not something that should have side-effects so maybe OK.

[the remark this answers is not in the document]
cpwright: I added the macro, convincing myself because we had an LHS and simple usage it should be fine.  You just can't be sloppy with it and use a value with side-effects.

src/main/c/jpy_jtype.c
cpwright: In the simpler cases, I decided to totally eliminate the slice, which gets rid of a real reference.  The getitem is a borrowed reference, so we don't need to worry about it.  In the complicated object case, I had to add additional decref calls.

devinrsmith:
Note: there is still work being done on this, as it has been causing some issues after https://github.com/deephaven/deephaven-core/pull/2105

devinrsmith:
Superseded by https://github.com/jpy-consortium/jpy/pull/45

PR #48, opened 2022-05-17
Release the GIL while in Java, with a special method to deliberately acquire it

devinrsmith:
There is one exceptional case to deal with, allowing Java to hold and
keep the GIL when it wants it, such as for cleanup work. Now Java can
grab the GIL if it needs it, without a full trip into python and back
again.


jmao-denver (approved):
LGTM

niloc132 (approved):

PR #53, opened 2022-05-31
Add x86 builds

devinrsmith:


devinrsmith:
Likely not a high priority to build on x86 (32-bit). Closing for now.

PR #85, opened 2022-08-26
Support creation of zero-sized primitive Java arrays

### LINKED ISSUE #84: Jpy should be able to create zero-sized arrays
## Steps to reproduce
Consider this snippet of code, which will call Java's `PrintWriter.write(byte[])` with an empty array:

```
import jpyutil
jpyutil.init_jvm()

import jpy
j_sys = jpy.get_type('java.lang.System')
j_sys.out.write(b'')
```

## Expected behavior
No error, array is passed to Java.

## Actual behavior
```
  File "<stdin>", line 1, in <module>
ValueError: illegal buffer argument: not a positive item count: 0
```

---

This use case is pretty contrived, but use cases could exist where it makes sense to support empty arrays passing from python to Java.

niloc132:
Fixes #84

### DIFF
--- a/src/main/c/jpy_jtype.c
+++ b/src/main/c/jpy_jtype.c
@@ -1938,13 +1938,6 @@ int JType_ConvertVarArgPyArgToJObjectArg(JNIEnv* jenv, JPy_ParamDescriptor* para
             }
 
             itemCount = pyBuffer->len / pyBuffer->itemsize;
-            if (itemCount < 0) {
-                PyBuffer_Release(pyBuffer);
-                PyMem_Del(pyBuffer);
-                JPy_DECREF(pyArg);
-                PyErr_Format(PyExc_ValueError, "illegal buffer argument: count must be non-negative: %ld", itemCount);
-                return -1;
-            }
 
             if (paramComponentType == JPy_JBoolean) {
                 jArray = (*jenv)->NewBooleanArray(jenv, itemCount);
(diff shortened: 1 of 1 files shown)


src/main/c/jpy_jtype.c
```diff
@@ -1938,11 +1938,11 @@ int JType_ConvertVarArgPyArgToJObjectArg(JNIEnv* jenv, JPy_ParamDescriptor* para
             }
 
             itemCount = pyBuffer->len / pyBuffer->itemsize;
-            if (itemCount <= 0) {
+            if (itemCount < 0) {
```
devinrsmith, line 1941: Assuming `pyBuffer` is correct, there shouldn't be any way to trigger this. I wonder if we should prefer to remove the blocks

niloc132: Aside from trying to avoid creating zero-length arrays (which apparently work correctly, but perhaps it would be better to only keep references to singleton empty arrays?), I can't imagine why this check would have existed.

It isn't the only `size < 0` check in the codebase, so it might still serve as a reasonable sanity check? I could easily go either way, will be guided by your preference.

devinrsmith: The majority of `< 0` checks are against return values where `-1` represents an exceptional state.

```c
        itemCount = PySequence_Length(pyArg);
        if (itemCount < 0) {
            return -1;
        }
```

ie, this returns `-1` if `pyArg` is *not* a sequence.

The one other extraneous check I could find is:

```c
    *size = PyUnicode_GET_SIZE(pyUnicode);
    if (*size < 0) {
        goto error;
    }
```

I don't see any cpython documentation that this is expected to ever trigger - the assumption is that `pyUnicode` is a unicode object; and the code base *does* make that check.

As such, I think it makes sense to remove these exception checking blocks.

devinrsmith (approved):

PR #88, opened 2022-09-23
Use valueOf() to box primitive values instead of creating new objects every time

rbasralian:


src/main/c/jpy_jtype.c
```diff
@@ -401,48 +401,59 @@ int JType_CreateJavaObject_2(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jcl
     return 0;
 }
 
+#define JType_CallStaticObjectMethodPrimitiveArgAndReturn(TARGET_TYPE, STATIC_METHOD_ID, VALUE, JOBJECT_PTR) \
```
devinrsmith, line 404: I think we want to give a better indication that this is a macro - I think we use all caps elsewhere.

src/main/c/jpy_jtype.c
```diff
@@ -401,48 +401,59 @@ int JType_CreateJavaObject_2(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jcl
...
+    Py_BEGIN_ALLOW_THREADS; \
+    *JOBJECT_PTR = (*jenv)->CallStaticObjectMethod(jenv, TARGET_TYPE, STATIC_METHOD_ID, VALUE); \
+    Py_END_ALLOW_THREADS; \
```
devinrsmith, line 407: It's not clear to me that we should / want to drop and then re-acquire the GIL (even thought the previous code to this _was_). In the other cases of CallStaticObjectMethod (and NewObject, etc) we are usually calling _arbitrary_ pieces of code - at least in the context of this patch, we are only calling short valueOf methods.

@niloc132 @jmao-denver may have perspective as well.

src/main/c/jpy_jtype.c
```diff
@@ -401,48 +401,59 @@ int JType_CreateJavaObject_2(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jcl
...
-    jvalue value;
+    jboolean value;
     if (PyBool_Check(pyArg) || JPy_IS_CLONG(pyArg)) {
-        value.z = JPy_AS_JBOOLEAN(pyArg);
+        value = JPy_AS_JBOOLEAN(pyArg);
     } else {
         return JType_PythonToJavaConversionError(type, pyArg);
     }
-    return JType_CreateJavaObject(jenv, type, pyArg, JPy_Boolean_JClass, JPy_Boolean_Init_MID, value, objectRef);
+    JType_CallStaticObjectMethodPrimitiveArgAndReturn(JPy_Boolean_JClass, JPy_Boolean_ValueOf_SMID, value, objectRef);
```
devinrsmith, line 423: Extra `;` (also in other places) given the macro defines it as well.

src/main/c/jpy_jtype.c
```diff
@@ -459,68 +470,66 @@ int JType_CreateJavaNumberFromPythonInt(JNIEnv* jenv, JPy_JType* type, PyObject*
 
     if (i != j) {
         value.j = j;
-        return JType_CreateJavaObject(jenv, type, pyArg, JPy_Long_JClass, JPy_Long_Init_MID, value, objectRef);
+        JType_CallStaticObjectMethodPrimitiveArgAndReturn(JPy_Long_JClass, JPy_Long_ValueOf_SMID, value.j, objectRef);
```
devinrsmith, line 473: I think `jvalue value` becomes unneccessary given pattern?

devinrsmith (changes requested):
I like it. Few small things.

src/main/c/jpy_jtype.c
rbasralian: renamed

src/main/c/jpy_jtype.c
rbasralian: Any Java allocation could theoretically lead to a lengthy GC so worth keeping that in mind

Is there any performance testing around jpy/any idea what the cost of dropping/reacquiring the GIL is?

src/main/c/jpy_jtype.c
```diff
@@ -457,70 +479,67 @@ int JType_CreateJavaNumberFromPythonInt(JNIEnv* jenv, JPy_JType* type, PyObject*
     s = (short) j;
     b = (signed char) j;
 
+    // always create new objects for long/int/short; any values assigned
+    // assigned to those types will be outside the default JVM cache
+    // range of [-128, 127].
     if (i != j) {
-        value.j = j;
-        return JType_CreateJavaObject(jenv, type, pyArg, JPy_Long_JClass, JPy_Long_Init_MID, value, objectRef);
+        JType_CREATE_JAVA_OBJECT_1_AND_RETURN(JPy_Long_JClass, JPy_Long_Init_MID, j, objectRef);
```
devinrsmith, line 486: So, it's not clear to me there will be any negligible performance difference if we keep doing valueOf. In fact, these constructors are *deprecated*. Not that I think they'll ever go away, but it might be worthwhile to stick w/ valueOf.

devinrsmith: I also don't know if it makes a difference, but valueOf is HotSpotIntrinsicCandidate (I don't know if constructors are or can be?).

src/main/c/jpy_jtype.c
devinrsmith: It's an interesting question - are allocation points also "safepoints"? I think "yes", but the real answer may be more nuanced than that.

That said, you are probably right and the way we do it now is probably the safest option.

src/main/c/jpy_jtype.c
rbasralian: Ok, thanks, good point. I changed it back. Doesn't make sense to use a deprecated code for performance reasons if there's no evidence it helps performance

src/main/c/jpy_jtype.c
rbasralian: I can change it, but I figured better this way because it's nice to "pretend it's a regular function" and because this is how other similar macros are used (e.g. `JPy_DELETE_LOCAL_REF`)

devinrsmith (approved):

Issue #93, opened 2022-10-31
Publish python 3.11 wheels


devinrsmith:

devinrsmith:
Looks like there may be some C layer changes we need to account for:

```
2022-10-31T20:58:35.2083924Z src/main/c/jni/org_jpy_PyLib.c: In function ‘format_python_traceback’:
2022-10-31T20:58:35.2084495Z src/main/c/jni/org_jpy_PyLib.c:2743:25: error: dereferencing pointer to incomplete type ‘PyFrameObject’ {aka ‘struct _frame’}
2022-10-31T20:58:35.2084988Z  2743 |             tb->tb_frame->f_code->co_filename != last_file ||
2022-10-31T20:58:35.2085253Z       |                         ^~
```

devinrsmith:
See notes in https://docs.python.org/3.11/whatsnew/3.11.html#whatsnew311-c-api-porting, 

> The [PyFrameObject](https://docs.python.org/3.11/c-api/frame.html#c.PyFrameObject) structure members have been removed from the public C API.

devinrsmith:
Depends on #95

devinrsmith:
3.11 wheels have been released.

PR #94, opened 2022-10-31
Produce python 3.11 artifacts for 0.12.0

devinrsmith:


devinrsmith:
Will need a version bump due to c api breakage, see #93

PR #99, opened 2023-03-24
Check for exception in getInt/Long/DoubleValue()

### LINKED ISSUE #98: Uncaught exception in PyObject getIntValue() showes up in the next PyListWrapper.size() call
In Python,
```
def foo():
    return [[], None]
```
In Java, after calling foo() and let 'obj' receive the return value as PyObject
```
obj.get(1).getIntValue();
obj.get(0).asList().size()
```
would throw an exception
```
java.lang.RuntimeException: Error in Python interpreter:
Type: <class 'SystemError'>
Value: <method-wrapper '__len__' of list object at 0x158099dc0> returned a result with an exception set
Line: <not available>
Namespace: <not available>
File: <not available>

        at org.jpy.PyLib.callAndReturnObject(Native Method)
        at org.jpy.PyObject.callMethod(PyObject.java:432)
        at org.jpy.PyListWrapper.size(PyListWrapper.java:36)
```

jmao-denver:
Fixes #98

### DIFF
--- a/src/main/c/jni/org_jpy_PyLib.c
+++ b/src/main/c/jni/org_jpy_PyLib.c
@@ -1196,6 +1196,11 @@ JNIEXPORT jint JNICALL Java_org_jpy_PyLib_getIntValue
 
     pyObject = (PyObject*) objId;
     value = (jint) JPy_AS_CLONG(pyObject);
+    // Note: we are not handling the overflow case, but there might be a need to have getIntValueSafe() in the future
+    if (value == -1 && PyErr_Occurred()) {
+        JPy_DIAG_PRINT(JPy_DIAG_F_ALL, "Java_org_jpy_PyLib_getIntValue: error: failed to convert Python object to Java int\n");
+        PyLib_HandlePythonException(jenv);
+    }
 
     JPy_END_GIL_STATE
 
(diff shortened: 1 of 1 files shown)


src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -1217,7 +1221,10 @@ JNIEXPORT jlong JNICALL Java_org_jpy_PyLib_getLongValue
 
     pyObject = (PyObject*) objId;
     value = JPy_AS_CLONG(pyObject);
-
+    if (value == -1 && PyErr_Occurred()) {
+        JPy_DIAG_PRINT(JPy_DIAG_F_ALL, "Java_org_jpy_PyLib_getLongValue: error: failed to convert Python object to Java long\n");
+        PyLib_HandlePythonException(jenv);
+    }
```
devinrsmith, line 1227: I think we want to consider other exceptional cases - for example, when the value overflows, it throws an exception instead of returning an error value: https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -1196,6 +1196,10 @@ JNIEXPORT jint JNICALL Java_org_jpy_PyLib_getIntValue
 
     pyObject = (PyObject*) objId;
     value = (jint) JPy_AS_CLONG(pyObject);
+    if (value == -1 && PyErr_Occurred()) {
+        JPy_DIAG_PRINT(JPy_DIAG_F_ALL, "Java_org_jpy_PyLib_getIntValue: error: failed to convert Python object to Java int\n");
+        PyLib_HandlePythonException(jenv);
+    }
```
devinrsmith, line 1202: We should probably try and match overflow semantics with whatever getLongValue does. The code as-is will just return the lowest 32-bits, which may, or may not, be what we want...

src/main/c/jni/org_jpy_PyLib.c
jmao-denver: I don't think the Overflow exception and returning -1 are mutually exclusive. That condition check is a common pattern even in the cpython source.

devinrsmith (approved):

PR #105, opened 2023-07-05
Remove try with resource in PyDictWrapper.values()

### LINKED ISSUE #104: PyDictWrapper.values() incorrectly close the underlying PyObject while it is still referenced.
Test code:
```
    public void simpleDict() {
        PyDictWrapper dict =
                PyObject.executeCode("{'mock':'yeah', 'ing':'yeah', 'bird':'yeah'}", PyInputMode.EXPRESSION).asDict();

       Assert.assertTrue(!dict.values().isEmpty());
        for (PyObject v : dict.values()) {
            Assert.assertEquals("yeah", v.toString());
        }
```
Result:
```
    java.lang.IllegalStateException: PyObjectState has already been taken
        at org.jpy.PyObjectState.borrowPointer(PyObjectState.java:28)
        at org.jpy.PyObject.getPointer(PyObject.java:165)
        at org.jpy.PyObject.callMethod(PyObject.java:432)
        at org.jpy.PyListWrapper.size(PyListWrapper.java:36)
        at org.jpy.PyListWrapper.isEmpty(PyListWrapper.java:43)
        at io.deephaven.jpy.integration.PyDictTest.simpleDict(PyDictTest.java:65)

```

jmao-denver:
Fixes #104

### DIFF
--- a/src/main/java/org/jpy/PyDictWrapper.java
+++ b/src/main/java/org/jpy/PyDictWrapper.java
@@ -174,9 +174,7 @@ public class PyDictWrapper implements Map<PyObject, PyObject>, AutoCloseable {
      */
     @Override
     public Collection<PyObject> values() {
-        try (final PyObject pyObj = PyLib.pyDictValues(this.pyObject.getPointer())) {
-            return pyObj.asList();
-        }
+        return PyLib.pyDictValues(this.pyObject.getPointer()).asList()
     }
 
     /**
--- a/src/main/java/org/jpy/PyDictWrapper.java
+++ b/src/main/java/org/jpy/PyDictWrapper.java
@@ -174,7 +174,7 @@ public class PyDictWrapper implements Map<PyObject, PyObject>, AutoCloseable {
      */
     @Override
     public Collection<PyObject> values() {
-        return PyLib.pyDictValues(this.pyObject.getPointer()).asList()
+        return PyLib.pyDictValues(this.pyObject.getPointer()).asList();
     }
 
     /**


jmao-denver:
> I think we need to audit all usages of asList and asDict, as it looks like there may be other callsites that use the same pattern.

I did a global search/verification and found no problem with other places that use the same pattern.

devinrsmith:
No, I see exactly the same issues w/ org.jpy.PyDictWrapper#keySet

devinrsmith:
I see exactly the same issues w/ org.jpy.PyDictWrapper#keySet; _maybe_ org.jpy.PyDictWrapper#entrySet (need to look into the semantics more, org.jpy.PyDictWrapper#get(java.lang.Object) claims a *borrowed* reference.

niloc132:
> I see exactly the same issues w/ org.jpy.PyDictWrapper#keySet; _maybe_ org.jpy.PyDictWrapper#entrySet (need to look into the semantics more, org.jpy.PyDictWrapper#get(java.lang.Object) claims a _borrowed_ reference.

I believe you're misreading (though I read it that way when I first looked at this PR): whereas PyDictWrapper.values returns `PyObject.asList()` (a `PyListWrapper` instance, wrapping the existing PyObject, and so the existing PyObject must not be closed/released), `PyDictWraper.keySet()` _copies_ the returned list into a new fresh LinkedHashSet (so the old one can and must be disposed of). The `entrySet()` implementation does the same, but washes it through a stream, collecting instances into a new LinkedHashSet rather than returning the `PyListWrapper` directly.

Both `keySet` and `entrySet` have the distinction that jpy offers no Set implementation, so a copy is necessary, and a copy means that we aren't providing a proper "view" of the underlying Map, which is technically wrong. On the other hand, `values` is returning a List instance that reflects the changes of the underlying dict/Map, which is correct, but it does mean that the instance can't be closed. Necessarily, the caller of `values()` must close the object  when done, but need not do so for `keySet` or `entrySet`.

`PyDictWrapper.get()` is a different case, but should behave like any other PyObject that is returned - the caller must close it when finished.

devinrsmith:
Good point. These are subtle issues.

PR #112, opened 2023-09-22
Add jpy.byte_buffer() function

### LINKED ISSUE #111: Add the ability to create Java DirectByteBuffer from Python objects that support the Buffer Protocol

jmao-denver:
Fixes #111 
Dependent on by https://github.com/deephaven/deephaven-core/pull/4936

1. byte_buffer() utility function to create a Java direct ByteBuffer wrapper that shares the underlying buffer of a Python buffer object (that implements the buffer protocol).
2. when calling a Java method that takes a ByteBuffer argument,  a Java ByteBuffer wrapper can be passed in.
3. when calling a Java method with the last argument being a variadic one of Java ByteBuffer type, a sequence of Java ByteBuffer wrappers made from calling byte_buffer() util function can be accepted.

Note, Java methods that receive ByteBuffer arguments are considered to be `borrowing` these buffers, not `owning` them, and they are only guaranteed to be safe to access for the duration of the methods. So if these Python buffers are to be used in Java beyond these method calls, it falls to the users to keep the Python objects from being GCed after the Java method calls finish or make copies of the buffers before returning from the method calls.

```
import jpyutil
jpyutil.init_jvm()
import jpy

def check_jbb(jbb):
    print(jbb.toString())
    print(f"isReadOnly: {jbb.isReadOnly()}")
    print("before scanning...")
    print(f"remaining: {jbb.remaining()}")
    print(f"position: {jbb.position()}")
    for i in range(jbb.remaining()):
        print(i, jbb.get())
    print("after scanning...")
    print(jbb.toString())
    print(f"remaining: {jbb.remaining()}")
    print(f"position: {jbb.position()}")


ba = b'abc'
jbb = jpy.byte_buffer(ba)
check_jbb(jbb)

import pyarrow as pa

data = [
    pa.array([1, 2, 3, 4]),
    pa.array(['foo', 'bar', 'baz', None]),
    pa.array([True, None, False, True])
]


batch = pa.record_batch(data, names=['f0', 'f1', 'f2'])
sink = pa.BufferOutputStream()
with pa.ipc.new_stream(sink, batch.schema) as writer:
   for i in range(5):
      writer.write_batch(batch)

buf = sink.getvalue()
jbb = jpy.byte_buffer(buf)
check_jbb(jbb)

buf = batch.schema.serialize()
jbb = jpy.byte_buffer(buf)
check_jbb(jbb)

```

### DIFF
--- a/src/main/c/jpy_jbyte_buffer.c
+++ b/src/main/c/jpy_jbyte_buffer.c
@@ -20,5 +20,5 @@
 #include "jpy_jbyte_buffer.h"
 
 /* 
- * This file for now is just a place-holder for future JPy_ByteBufferWrapper specific functions.
+ * This file for now is just a place-holder for future JPy_JByteBufferObj specific functions.
  */
--- a/src/main/c/jpy_jbyte_buffer.c
+++ b/src/main/c/jpy_jbyte_buffer.c
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2023 JPY-CONSORTIUM Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jpy_module.h"
+#include "jpy_diag.h"
+#include "jpy_jarray.h"
+#include "jpy_jbyte_buffer.h"
+
+/* This file for now is just a place-holder for future JPy_ByteBufferWrapper specific functions.
+ */
--- a/src/main/c/jpy_jbyte_buffer.c
+++ b/src/main/c/jpy_jbyte_buffer.c
@@ -19,5 +19,6 @@
 #include "jpy_jarray.h"
 #include "jpy_jbyte_buffer.h"
 
-/* This file for now is just a place-holder for future JPy_ByteBufferWrapper specific functions.
+/* 
+ * This file for now is just a place-holder for future JPy_ByteBufferWrapper specific functions.
  */
--- a/src/main/c/jpy_jbyte_buffer.h
+++ b/src/main/c/jpy_jbyte_buffer.h
@@ -26,7 +26,7 @@ extern "C" {
 /**
  * The Java ByteBuffer representation in Python.
  *
- * IMPORTANT: JPy_JByteBufferWrapper must only differ from the JPy_JObj structure by the 'obj' member
+ * IMPORTANT: JPy_JByteBufferWrapper must only differ from the JPy_JObj structure by the 'pyBuffer' member
  * since we use the same basic type, name JPy_JType for it. DON'T ever change member positions!
  * @see JPy_JObj
  */
--- a/src/main/c/jpy_jbyte_buffer.h
+++ b/src/main/c/jpy_jbyte_buffer.h
@@ -26,17 +26,17 @@ extern "C" {
 /**
  * The Java ByteBuffer representation in Python.
  *
- * IMPORTANT: JPy_JByteBufferWrapper must only differ from the JPy_JObj structure by the 'pyBuffer' member
+ * IMPORTANT: JPy_JByteBufferObj must only differ from the JPy_JObj structure by the 'pyBuffer' member
  * since we use the same basic type, name JPy_JType for it. DON'T ever change member positions!
  * @see JPy_JObj
  */
-typedef struct JPy_JByteBufferWrapper
+typedef struct JPy_JByteBufferObj
 {
     PyObject_HEAD
     jobject objectRef;
     Py_buffer *pyBuffer;
 }
-JPy_JByteBufferWrapper;
+JPy_JByteBufferObj;
 
 #ifdef __cplusplus
 }  /* extern "C" */
--- a/src/main/c/jpy_jbyte_buffer.h
+++ b/src/main/c/jpy_jbyte_buffer.h
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2023 JPY-CONSORTIUM Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JPY_JBYTE_BUFFER_H
+#define JPY_JBYTE_BUFFER_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "jpy_compat.h"
+
+/**
+ * The Java ByteBuffer representation in Python.
+ *
+ * IMPORTANT: JPy_JByteBufferWrapper must only differ from the JPy_JObj structure by the 'obj' member
+ * since we use the same basic type, name JPy_JType for it. DON'T ever change member positions!
+ * @see JPy_JObj
+ */
+typedef struct JPy_JByteBufferWrapper
+{
+    PyObject_HEAD
+    jobject objectRef;
+    Py_buffer *pyBuffer;
+}
+JPy_JByteBufferWrapper;
+
+#ifdef __cplusplus
+}  /* extern "C" */
+#endif
+#endif /* !JPY_JBYTE_BUFFER_H */
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -64,9 +64,13 @@ PyObject* JObj_FromType(JNIEnv* jenv, JPy_JType* type, jobject objectRef)
         array = (JPy_JArray*) obj;
         array->bufferExportCount = 0;
         array->buf = NULL;
+    } else if ((*jenv)->IsInstanceOf(jenv, objectRef, JPy_ByteBuffer_JClass)) {
+        JPy_JByteBufferWrapper *byteBufferWrapper = (JPy_JByteBufferWrapper *) obj;
+        byteBufferWrapper->pyBuffer = NULL;
     }
 
-    // we check the type translations dictionary for a callable for this java type name,
+
+// we check the type translations dictionary for a callable for this java type name,
     // and apply the returned callable to the wrapped object
     callable = PyDict_GetItemString(JPy_Type_Translations, type->javaName);
     if (callable != NULL) {
@@ -172,6 +176,8 @@ void JObj_dealloc(JPy_JObj* self)
     JNIEnv* jenv;
     JPy_JType* jtype;
 
+    jenv = JPy_GetJNIEnv();
+
     JPy_DIAG_PRINT(JPy_DIAG_F_MEM, "JObj_dealloc: releasing instance of %s, self->objectRef=%p\n", Py_TYPE(self)->tp_name, self->objectRef);
 
     jtype = (JPy_JType *)Py_TYPE(self);
@@ -182,14 +188,14 @@ void JObj_dealloc(JPy_JObj* self)
         if (array->buf != NULL) {
             JArray_ReleaseJavaArrayElements(array, array->javaType);
         }
-    } else if (jtype == JPy_JByteBuffer) {
-        JPy_JByteBufferWrapper* byteBufferWrapper;
-        byteBufferWrapper = (JPy_JByteBufferWrapper *) self;
-        PyBuffer_Release(byteBufferWrapper->pyBuffer);
-        PyMem_Free(byteBufferWrapper->pyBuffer);
+    } else if ((*jenv)->IsInstanceOf(jenv, self->objectRef, JPy_ByteBuffer_JClass)) {
+        JPy_JByteBufferWrapper *byteBufferWrapper = (JPy_JByteBufferWrapper *) self;
+        if (byteBufferWrapper->pyBuffer != NULL) {
+            PyBuffer_Release(byteBufferWrapper->pyBuffer);
+            PyMem_Free(byteBufferWrapper->pyBuffer);
+        }
     }
 
-    jenv = JPy_GetJNIEnv();
     if (jenv != NULL) {
         if (self->objectRef != NULL) {
             (*jenv)->DeleteGlobalRef(jenv, self->objectRef);
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -712,7 +712,6 @@ int JType_InitSlots(JPy_JType* type)
     PyTypeObject* typeObj;
     jboolean isArray;
     jboolean isPrimitiveArray;
-    jboolean isByteBuffer;
 
     isArray = type->componentType != NULL;
     isPrimitiveArray = isArray && type->componentType->isPrimitive;
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -20,6 +20,7 @@
 #include "jpy_module.h"
 #include "jpy_diag.h"
 #include "jpy_jarray.h"
+#include "jpy_jbyte_buffer.h"
 #include "jpy_jtype.h"
 #include "jpy_jobj.h"
 #include "jpy_jmethod.h"
@@ -181,8 +182,12 @@ void JObj_dealloc(JPy_JObj* self)
         if (array->buf != NULL) {
             JArray_ReleaseJavaArrayElements(array, array->javaType);
         }
-  
-    }    
+    } else if (jtype == JPy_JByteBuffer) {
+        JPy_JByteBufferWrapper* byteBufferWrapper;
+        byteBufferWrapper = (JPy_JByteBufferWrapper *) self;
+        PyBuffer_Release(byteBufferWrapper->pyBuffer);
+        PyMem_Free(byteBufferWrapper->pyBuffer);
+    }
 
     jenv = JPy_GetJNIEnv();
     if (jenv != NULL) {
@@ -707,9 +712,11 @@ int JType_InitSlots(JPy_JType* type)
     PyTypeObject* typeObj;
     jboolean isArray;
     jboolean isPrimitiveArray;
+    jboolean isByteBuffer;
 
     isArray = type->componentType != NULL;
     isPrimitiveArray = isArray && type->componentType->isPrimitive;
+    isByteBuffer = type == JPy_JByteBuffer;
 
     typeObj = JTYPE_AS_PYTYPE(type);
 
@@ -727,7 +734,7 @@ int JType_InitSlots(JPy_JType* type)
     //Py_SET_TYPE(type, &JType_Type);
     //Py_SET_SIZE(type, sizeof (JPy_JType));
 
-    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : sizeof (JPy_JObj);
+    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : (isByteBuffer ? sizeof(JPy_JByteBufferWrapper) : sizeof (JPy_JObj));
     typeObj->tp_itemsize = 0;
     typeObj->tp_base = type->superType != NULL ? JTYPE_AS_PYTYPE(type->superType) : &JType_Type;
     //typeObj->tp_base = (PyTypeObject*) type->superType;
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -64,7 +64,7 @@ PyObject* JObj_FromType(JNIEnv* jenv, JPy_JType* type, jobject objectRef)
         array = (JPy_JArray*) obj;
         array->bufferExportCount = 0;
         array->buf = NULL;
-    } else if ((*jenv)->IsInstanceOf(jenv, objectRef, JPy_ByteBuffer_JClass)) {
+    } else if (JByteBuffer_Check(type)) {
         JPy_JByteBufferWrapper *byteBufferWrapper;
 
         byteBufferWrapper = (JPy_JByteBufferWrapper *) obj;
@@ -178,8 +178,6 @@ void JObj_dealloc(JPy_JObj* self)
     JNIEnv* jenv;
     JPy_JType* jtype;
 
-    jenv = JPy_GetJNIEnv();
-
     JPy_DIAG_PRINT(JPy_DIAG_F_MEM, "JObj_dealloc: releasing instance of %s, self->objectRef=%p\n", Py_TYPE(self)->tp_name, self->objectRef);
 
     jtype = (JPy_JType *)Py_TYPE(self);
@@ -190,7 +188,7 @@ void JObj_dealloc(JPy_JObj* self)
         if (array->buf != NULL) {
             JArray_ReleaseJavaArrayElements(array, array->javaType);
         }
-    } else if ((*jenv)->IsInstanceOf(jenv, self->objectRef, JPy_ByteBuffer_JClass)) {
+    } else if (JByteBuffer_Check(jtype)) {
         JPy_JByteBufferWrapper *byteBufferWrapper;
         byteBufferWrapper = (JPy_JByteBufferWrapper *) self;
         if (byteBufferWrapper->pyBuffer != NULL) {
@@ -199,6 +197,7 @@ void JObj_dealloc(JPy_JObj* self)
         }
     }
 
+    jenv = JPy_GetJNIEnv();
     if (jenv != NULL) {
         if (self->objectRef != NULL) {
             (*jenv)->DeleteGlobalRef(jenv, self->objectRef);
@@ -743,7 +742,7 @@ int JType_InitSlots(JPy_JType* type)
 
     if (isPrimitiveArray) {
         typeObj->tp_basicsize = sizeof(JPy_JArray);
-    } else if (type == JPy_JByteBuffer) {
+    } else if (JByteBuffer_Check(type)) {
         typeObj->tp_basicsize = sizeof(JPy_JByteBufferWrapper);
     } else {
         typeObj->tp_basicsize = sizeof(JPy_JObj);
@@ -842,3 +841,13 @@ int JType_Check(PyObject* arg)
     return PyType_Check(arg) && JPY_IS_JTYPE(arg);
 }
 
+int JByteBuffer_Check(JPy_JType* type) {
+    while (type != NULL) {
+        if (type == JPy_JByteBuffer || strcmp(type->javaName, "java.nio.ByteBuffer") == 0) {
+            JPy_DIAG_PRINT(JPy_DIAG_F_TYPE, "JByteBuffer_Check: java ByteBuffer or its sub-type (%s) found.\n", type->javaName);
+            return -1;
+        }
+        type = type->superType;
+    }
+    return 0;
+}
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -716,7 +716,6 @@ int JType_InitSlots(JPy_JType* type)
 
     isArray = type->componentType != NULL;
     isPrimitiveArray = isArray && type->componentType->isPrimitive;
-    isByteBuffer = type == JPy_JByteBuffer;
 
     typeObj = JTYPE_AS_PYTYPE(type);
 
@@ -734,7 +733,13 @@ int JType_InitSlots(JPy_JType* type)
     //Py_SET_TYPE(type, &JType_Type);
     //Py_SET_SIZE(type, sizeof (JPy_JType));
 
-    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : (isByteBuffer ? sizeof(JPy_JByteBufferWrapper) : sizeof (JPy_JObj));
+    if (isPrimitiveArray) {
+        typeObj->tp_basicsize = sizeof(JPy_JArray);
+    } else if (type == JPy_JByteBuffer) {
+        typeObj->tp_basicsize = sizeof(JPy_JByteBufferWrapper);
+    } else {
+        typeObj->tp_basicsize = sizeof(JPy_JObj);
+    }
     typeObj->tp_itemsize = 0;
     typeObj->tp_base = type->superType != NULL ? JTYPE_AS_PYTYPE(type->superType) : &JType_Type;
     //typeObj->tp_base = (PyTypeObject*) type->superType;
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -65,10 +65,10 @@ PyObject* JObj_FromType(JNIEnv* jenv, JPy_JType* type, jobject objectRef)
         array->bufferExportCount = 0;
         array->buf = NULL;
     } else if (JByteBuffer_Check(type)) {
-        JPy_JByteBufferWrapper *byteBufferWrapper;
+        JPy_JByteBufferObj *byteBuffer;
 
-        byteBufferWrapper = (JPy_JByteBufferWrapper *) obj;
-        byteBufferWrapper->pyBuffer = NULL;
+        byteBuffer = (JPy_JByteBufferObj *) obj;
+        byteBuffer->pyBuffer = NULL;
     }
 
 
@@ -189,11 +189,11 @@ void JObj_dealloc(JPy_JObj* self)
             JArray_ReleaseJavaArrayElements(array, array->javaType);
         }
     } else if (JByteBuffer_Check(jtype)) {
-        JPy_JByteBufferWrapper *byteBufferWrapper;
-        byteBufferWrapper = (JPy_JByteBufferWrapper *) self;
-        if (byteBufferWrapper->pyBuffer != NULL) {
-            PyBuffer_Release(byteBufferWrapper->pyBuffer);
-            PyMem_Free(byteBufferWrapper->pyBuffer);
+        JPy_JByteBufferObj *byteBuffer;
+        byteBuffer = (JPy_JByteBufferObj *) self;
+        if (byteBuffer->pyBuffer != NULL) {
+            PyBuffer_Release(byteBuffer->pyBuffer);
+            PyMem_Free(byteBuffer->pyBuffer);
         }
     }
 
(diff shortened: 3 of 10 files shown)


src/main/c/jpy_jobj.c
```diff
@@ -727,7 +734,7 @@ int JType_InitSlots(JPy_JType* type)
     //Py_SET_TYPE(type, &JType_Type);
     //Py_SET_SIZE(type, sizeof (JPy_JType));
 
-    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : sizeof (JPy_JObj);
+    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : (isByteBuffer ? sizeof(JPy_JByteBufferWrapper) : sizeof (JPy_JObj));
```
cpwright, line 737: There are enough ternaries here that a function or if statement might make sense instead.  It is borderline with 3, so not strictly necessary.

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,53 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+        return NULL;
+    }
+
+    pyBuffer = (Py_buffer *)PyMem_Malloc(sizeof(Py_buffer));
+    if (pyBuffer == NULL) {
+        return PyErr_NoMemory();
+    }
+
+    if (PyObject_GetBuffer(pyObj, pyBuffer, PyBUF_SIMPLE | PyBUF_C_CONTIGUOUS) != 0) {
+        PyErr_SetString(PyExc_ValueError, "byte_buffer: the Python object failed to return a contiguous buffer.");
```
cpwright, line 693: did you leak pyBuffer here?

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,53 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+        return NULL;
+    }
+
+    byteBufferRef = (*jenv)->NewDirectByteBuffer(jenv, pyBuffer->buf, pyBuffer->len);
+    if (byteBufferRef == NULL) {
+        return PyErr_NoMemory();
```
cpwright, line 699: and here?

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,53 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+    }
+
+    newPyObj = JObj_New(jenv, byteBufferRef);
+    if (newPyObj == NULL) {
+        return NULL;
```
cpwright, line 704: byteBufferRef leaked on error

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,53 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+    }
+    byteBufferWrapper = (JPy_JByteBufferWrapper *) newPyObj;
+    byteBufferWrapper->pyBuffer = pyBuffer;
+    return (PyObject *)byteBufferWrapper;
```
cpwright, line 708: Do you need a destructor to clean up the direct buffer expicitly?

jmao-denver: It is handled by ```void JObj_dealloc``` in jpy_jobj.c.

niloc132 (commented):
Notes from call:
 * I like that we have a helper for this, it could be handy for explicit manipulation of the java type from python
 * Please consider tweaking the python->java call code so that if the matching method has a ByteBuffer argument, and the provided argument is apparently a python buffer, we automatically wrap in this way. 
 * Likewise, when Java calls python with a direct bytebuffer, consider automatically providing that to python as a buffer rather than assuming that the python being called already knows how to handle it. Non-direct buffers probably just have to be passed as-is.
 * Handling readonly should be as simple as calling .asReadOnlyBuffer() on the java ByteBuffer instance - this shouldn't copy, just provide a readonly view for Java, still leaving the instance as a direct bytebuffer.

src/main/c/jpy_module.c
devinrsmith: I don't see this as being resolved as of your latest commit.

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,56 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+
+    if (PyObject_GetBuffer(pyObj, pyBuffer, PyBUF_SIMPLE | PyBUF_C_CONTIGUOUS) != 0) {
+        PyErr_SetString(PyExc_ValueError, "byte_buffer: the Python object failed to return a contiguous buffer.");
+        PyMem_Free(pyBuffer);
+        return NULL;
+    }
+
+    byteBufferRef = (*jenv)->NewDirectByteBuffer(jenv, pyBuffer->buf, pyBuffer->len);
+    if (byteBufferRef == NULL) {
+        PyMem_Free(pyBuffer);
```
devinrsmith, line 700: I'm wondering, do we need to PyBuffer_Release in addition to or instead of PyMem_Free?

> Successful calls to [PyObject_GetBuffer()](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer) must be paired with calls to [PyBuffer_Release()](https://docs.python.org/3/c-api/buffer.html#c.PyBuffer_Release)

jmao-denver: Yes, you are right.

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,56 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+    if (!PyArg_ParseTuple(args, "O:byte_buffer", &pyObj)) {
+        return NULL;
+    }
+
+    if (PyObject_CheckBuffer(pyObj) == 0) {
+        PyErr_SetString(PyExc_ValueError, "byte_buffer: argument 1 must be a Python object that supports the buffer protocol");
+        return NULL;
+    }
+
+    pyBuffer = (Py_buffer *)PyMem_Malloc(sizeof(Py_buffer));
```
devinrsmith, line 687: I very rusty w/ C; but this is allocating us a raw pointer that we can use w/ PyObject_GetBuffer, yes?

Does it make sense to use slightly higher level interfaces, ie https://docs.python.org/3/c-api/memoryview.html#c.PyMemoryView_GetContiguous?

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,58 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+        return PyErr_NoMemory();
+    }
+
+    if (PyObject_GetBuffer(pyObj, pyBuffer, PyBUF_SIMPLE | PyBUF_C_CONTIGUOUS) != 0) {
+        PyErr_SetString(PyExc_ValueError, "byte_buffer: the Python object failed to return a contiguous buffer.");
+        PyMem_Free(pyBuffer);
+        return NULL;
+    }
+
+    byteBufferRef = (*jenv)->NewDirectByteBuffer(jenv, pyBuffer->buf, pyBuffer->len);
```
devinrsmith, line 698: We can call `java.nio.ByteBuffer#asReadOnlyBuffer` after this if we want / need read-only.

src/main/c/jpy_jobj.c
```diff
@@ -707,6 +712,7 @@ int JType_InitSlots(JPy_JType* type)
     PyTypeObject* typeObj;
     jboolean isArray;
     jboolean isPrimitiveArray;
+    jboolean isByteBuffer;
```
devinrsmith, line 715: unused?

devinrsmith (commented):
This is a fun PR. I generally like the ideas.

There's one big gotcha: this assumes that python will keep the reference alive as long as java needs it, and there is no way for java to let python know if it still needs the reference in the current approach.

For example:

```python
someJavaMethod(jpy.byte_buffer(some_buffer_obj))
```

will get a buffer, create a py/java ByteBuffer, pass it to someJavaMethod, and upon return, delete the python object, release the buffer, and tell jni to delete (decrement) the ByteBuffer; but java might still have a reference to the ByteBuffer, which now points to potentially invalid memory.

We _may_ be okay with this limitation, as long as we are very clear. I don't know if there is an easy workaround. My initial thoughts go to a jpy-owned java class that more closely models the buffer protocol; where the java object owns a PyObject (memoryview probably) to ensure it stays alive for as long as java has a reference to it.

Looking deeper though, there may actually be a better solution:

```java
    // Invoked to construct a direct ByteBuffer referring to the block of
    // memory. A given arbitrary object may also be attached to the buffer.
    //
    DirectByteBuffer(long addr, int cap, Object ob) {
        super(-1, 0, cap, cap);
        address = addr;
        cleaner = null;
        att = ob;
    }


    // Invoked only by JNI: NewDirectByteBuffer(void*, long)
    //
    private DirectByteBuffer(long addr, int cap) {
        super(-1, 0, cap, cap);
        address = addr;
        cleaner = null;
        att = null;
    }
```

```java
    // An object attached to this buffer. If this buffer is a view of another
    // buffer then we use this field to keep a reference to that buffer to
    // ensure that its memory isn't freed before we are done with it.
    private final Object att;
```

This comment on `att` is exactly what we want. If it's a pointer to our python obj repr of ByteBuffer (or if we use an auxilary memoryview python obj), then I _think_ that solves the problem.

I don't know exactly how to get at it yet, but there must be a way. `jdk.internal.misc.JavaNioAccess#newDirectByteBuffer`

src/main/c/jpy_jbyte_buffer.c
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+#include "jpy_module.h"
+#include "jpy_diag.h"
+#include "jpy_jarray.h"
+#include "jpy_jbyte_buffer.h"
+
+/* This file for now is just a place-holder for future JPy_ByteBufferWrapper specific functions.
+ */
```
chipkent, line 23:
```suggestion
/* 
 * This file for now is just a place-holder for future JPy_ByteBufferWrapper specific functions.
 */
```

devinrsmith:
Another small wrinkle for the future; java 21 has long for capacity.

src/main/c/jpy_module.c
```diff
@@ -660,6 +667,58 @@ PyObject* JPy_array(PyObject* self, PyObject* args)
...
+
+    pyBuffer = (Py_buffer *)PyMem_Malloc(sizeof(Py_buffer));
+    if (pyBuffer == NULL) {
+        return PyErr_NoMemory();
+    }
+
+    if (PyObject_GetBuffer(pyObj, pyBuffer, PyBUF_SIMPLE | PyBUF_C_CONTIGUOUS) != 0) {
+        PyErr_SetString(PyExc_ValueError, "byte_buffer: the Python object failed to return a contiguous buffer.");
+        PyMem_Free(pyBuffer);
+        return NULL;
```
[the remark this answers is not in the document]
jmao-denver: We should return Null, the calling function checks if an exception is raised.

src/main/c/jpy_jtype.c
```diff
@@ -1652,6 +1654,90 @@ int JType_ConvertPyArgToJStringArg(JNIEnv* jenv, JPy_ParamDescriptor* paramDescr
...
+        return NULL;
+    }
+
+    if (PyObject_GetBuffer(pyObj, pyBuffer, PyBUF_SIMPLE | PyBUF_C_CONTIGUOUS) != 0) {
+        PyErr_SetString(PyExc_ValueError, "JType_CreateJavaByteBufferWrapper: the Python object failed to return a contiguous buffer.");
+        PyMem_Free(pyBuffer);
+        return NULL;
+    }
+
+    tmpByteBufferRef = (*jenv)->NewDirectByteBuffer(jenv, pyBuffer->buf, pyBuffer->len);
```
devinrsmith, line 1688: I'm _very_ concerned about creating a byte buffer in this way - it's possible that the PyObject for the ByteBuffer goes away, but that the ByteBuffer is still accessed via java. It's unfortunate the JNI doesn't provider methods for calling `DirectByteBuffer(long addr, int cap, Object ob)` directly, but `ob` here is exactly what we need:

```java
    // An object attached to this buffer. If this buffer is a view of another
    // buffer then we use this field to keep a reference to that buffer to
    // ensure that its memory isn't freed before we are done with it.
    private final Object att;
```

src/main/c/jpy_module.c
```diff
@@ -63,6 +65,9 @@ static PyMethodDef JPy_Functions[] = {
                     "array(name, init) - Return a new Java array of given Java type (type name or type object) and initializer (array length or sequence). "
                     "Possible primitive types are 'boolean', 'byte', 'char', 'short', 'int', 'long', 'float', and 'double'."},
 
+    {"byte_buffer", JPy_byte_buffer, METH_VARARGS,
```
devinrsmith, line 68: If we're unable to implement the proper java lifecycle attachment, at a minimum we need to strongly document the limitations of the method - "the python ByteBuffer object must live at least as long as the java ByteBuffer". Additionally, I would want a better name like `byte_buffer_unsafe` to further warn against misuse.

src/main/c/jpy_jtype.c
```diff
@@ -1652,6 +1654,90 @@ int JType_ConvertPyArgToJStringArg(JNIEnv* jenv, JPy_ParamDescriptor* paramDescr
     return JPy_AsJString(jenv, pyArg, &value->l);
 }
 
+int JType_MatchPyArgAsJByteBufferParam(JNIEnv* jenv, JPy_ParamDescriptor* paramDescriptor, PyObject* pyArg)
```
devinrsmith, line 1657: Given caveats about the limitations as currently implemented, I'm against any sort of auto-conversion.

devinrsmith (commented):
I've got strong reservations about the current impl. I _think_ there are ways that we could cleanly fix it.

Also, I think there is a leak:

```python
import jpyutil
jpyutil.init_jvm()

import ctypes
import gc
import jpy


def get_refcount(obj_id):
  return ctypes.c_long.from_address(obj_id).value


def garbage_collect() -> None:
    gc.collect()
    _j_system = jpy.get_type("java.lang.System")
    _j_system.gc()
    _j_system.gc()


x = b'these are some bytes'
x_id = id(x)

print(f"get_refcount(x_id)={get_refcount(x_id)}")

bb = jpy.byte_buffer(x)

# this works correctly
# bb = memoryview(x)

print(f"get_refcount(x_id)={get_refcount(x_id)}")

del bb
garbage_collect()

print(f"get_refcount(x_id)={get_refcount(x_id)}")
```

results in

```
get_refcount(x_id)=1
get_refcount(x_id)=2
get_refcount(x_id)=2
```

src/main/c/jpy_module.c
jmao-denver: I am not sure adding the suffix `_unsafe` is necessary, it might even cause confusion in the future when we have a solution to the Java lifecycle issue because it is no longer true and removing it will be a breaking change.

src/main/c/jpy_jobj.c
```diff
@@ -181,10 +188,14 @@ void JObj_dealloc(JPy_JObj* self)
         if (array->buf != NULL) {
             JArray_ReleaseJavaArrayElements(array, array->javaType);
         }
-  
-    }    
+    } else if ((*jenv)->IsInstanceOf(jenv, self->objectRef, JPy_ByteBuffer_JClass)) {
+        JPy_JByteBufferWrapper *byteBufferWrapper = (JPy_JByteBufferWrapper *) self;
```
devinrsmith, line 192: Is it ever possible we'll have a plain JObj ByteBuffer? For example, what if we have:

```java
static ByteBuffer myBuffer() {
...
}
```

```python
my_byte_buffer = jpy.get_type('SomeClass').myBuffer()
del my_byte_buffer
```

Or, is that already handled and we expect `byteBufferWrapper->pyBuffer` to be null in that case?

What about if the signature was instead:

```java
static Buffer myBuffer() {
...
}
```

The python type is generic on `Buffer` right, but then the actual instance _may_ be ByteBuffer, and I suspect it won't be a JPy_JByteBufferWrapper type?

src/main/c/jpy_jtype.c
```diff
@@ -1652,6 +1653,81 @@ int JType_ConvertPyArgToJStringArg(JNIEnv* jenv, JPy_ParamDescriptor* paramDescr
...
+    if (byteBufferRef == NULL) {
+        PyBuffer_Release(pyBuffer);
+        PyMem_Free(pyBuffer);
+        JPy_DELETE_LOCAL_REF(tmpByteBufferRef);
+        PyErr_SetString(PyExc_RuntimeError, "jpy: internal error: failed to create a read-only ByteBuffer instance.");
+        return NULL;
+    }
+    JPy_DELETE_LOCAL_REF(tmpByteBufferRef);
+
+    PyObject *newPyObj = JObj_New(jenv, byteBufferRef);
```
devinrsmith, line 1701: I see how we are washing through `JObj_FromType` here; but it strikes me as a bit weird to not have a more specific construction call where we know we get a JPy_JByteBufferWrapper back... maybe `JByteBuffer_New`?

src/main/c/jpy_jobj.c
```diff
@@ -181,10 +188,14 @@ void JObj_dealloc(JPy_JObj* self)
         if (array->buf != NULL) {
             JArray_ReleaseJavaArrayElements(array, array->javaType);
         }
-  
-    }    
+    } else if ((*jenv)->IsInstanceOf(jenv, self->objectRef, JPy_ByteBuffer_JClass)) {
```
devinrsmith, line 191: Needing to rely on jenv for this seems wrong (and potentially incorrect behavior, see other comment); shouldn't we have enough info in the type to figure it out? If not, can we add enough info to the type?

I wonder if we should have some equivalent to `JObj_Check`, but for our specific types: `JByteBuffer_Check`?

src/main/c/jpy_jobj.c
```diff
@@ -63,9 +64,13 @@ PyObject* JObj_FromType(JNIEnv* jenv, JPy_JType* type, jobject objectRef)
         array = (JPy_JArray*) obj;
         array->bufferExportCount = 0;
         array->buf = NULL;
+    } else if ((*jenv)->IsInstanceOf(jenv, objectRef, JPy_ByteBuffer_JClass)) {
+        JPy_JByteBufferWrapper *byteBufferWrapper = (JPy_JByteBufferWrapper *) obj;
```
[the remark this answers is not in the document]
kosak: In the bad old days you were required to separate declaration and assignment. In modern C you are allowed to combine them.

I think the combined form is preferable unless you're adhering to someone else's strict style guide.

src/main/c/jpy_jtype.c
```diff
@@ -2451,6 +2527,9 @@ void JType_InitParamDescriptorFunctions(JPy_ParamDescriptor* paramDescriptor, jb
     } else if (paramType == JPy_JString) {
         paramDescriptor->MatchPyArg = JType_MatchPyArgAsJStringParam;
         paramDescriptor->ConvertPyArg = JType_ConvertPyArgToJStringArg;
+    } else if (paramType == JPy_JByteBuffer) {
```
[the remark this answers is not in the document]
jmao-denver: No, we don't actually need to explicitly handle JPy_JByteBuffer after I removed the Py_buffer auto conversion code. I decided to keep  the code here in anticipation that we might be able to use the direct byte buffer API to safely manage the lifecycle of the buffers. Removed.

src/main/c/jpy_jbyte_buffer.h
```diff
@@ -0,0 +1,44 @@
...
+#include "jpy_compat.h"
+
+/**
+ * The Java ByteBuffer representation in Python.
+ *
+ * IMPORTANT: JPy_JByteBufferWrapper must only differ from the JPy_JObj structure by the 'pyBuffer' member
+ * since we use the same basic type, name JPy_JType for it. DON'T ever change member positions!
+ * @see JPy_JObj
+ */
+typedef struct JPy_JByteBufferWrapper
```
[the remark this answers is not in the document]
jmao-denver: To follow the convention, `JPy_JByteBuffer` is the name of the Python type for ByteBuffer.  I think the more explicit name `JPy_JByteBufferWrapper` describes our intention better.

src/main/c/jpy_jobj.c
jmao-denver: Yes, the `pyBuffer` field will be NULL in that case. And it will be a `JPy_JByteBufferWrapper` instance.

src/main/c/jpy_jtype.c
jmao-denver: Having a `JByteBuffer_New` might work but I think this is also appropriate because we also need to handle the not-from-JNI ByteBuffer case which

src/main/c/jpy_jobj.c
jmao-denver: That's the first thing I tried. But the JNI-made direct byte buffer is of java.nio.DirectByteBufferR class and I need to check if it is assignable to ByteBuffer anyway.

src/main/c/jpy_jtype.c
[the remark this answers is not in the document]
jmao-denver: It is tested in https://github.com/deephaven/deephaven-core/pull/4936

src/main/c/jpy_jobj.c
```diff
@@ -63,9 +64,15 @@ PyObject* JObj_FromType(JNIEnv* jenv, JPy_JType* type, jobject objectRef)
         array = (JPy_JArray*) obj;
         array->bufferExportCount = 0;
         array->buf = NULL;
+    } else if ((*jenv)->IsInstanceOf(jenv, objectRef, JPy_ByteBuffer_JClass)) {
+        JPy_JByteBufferWrapper *byteBufferWrapper;
+
+        byteBufferWrapper = (JPy_JByteBufferWrapper *) obj;
+        byteBufferWrapper->pyBuffer = NULL;
```
devinrsmith, line 71: I think this is fundamentally unsafe. See comments later.

src/main/c/jpy_jobj.c
devinrsmith: I still think this is fundamentally unsafe. See comments later.

src/main/c/jpy_jobj.c
```diff
@@ -727,7 +741,13 @@ int JType_InitSlots(JPy_JType* type)
     //Py_SET_TYPE(type, &JType_Type);
     //Py_SET_SIZE(type, sizeof (JPy_JType));
 
-    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : sizeof (JPy_JObj);
+    if (isPrimitiveArray) {
+        typeObj->tp_basicsize = sizeof(JPy_JArray);
+    } else if (type == JPy_JByteBuffer) {
+        typeObj->tp_basicsize = sizeof(JPy_JByteBufferWrapper);
```
devinrsmith, line 747: We only allocate enough size for JPy_JByteBufferWrapper when the type is *explicitly* `type == JPy_JByteBuffer`, yet the other callsites _assume_ JPy_JByteBufferWrapper based on IsInstanceOf checks.  This is why I think we need more explicit data in the type to check if this is really a type that we've wrapped w/ the proper sizing, b/c otherwise we may operate on data we don't own.

devinrsmith: I'm running some tests to demonstrate the issue.

devinrsmith: I actually don't see this line being called ever, which makes me think there is something fundamentally wrong. I added some logging here, as well as additional logging out of tp_basicsize.

Here's the relevant lines:

```
$ grep JType_InitSlots log.txt | grep ByteBuffer
JType_InitSlots: typeObj=0x7fcb76ec2870, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="java.nio.ByteBuffer", typeObj->tp_base=0x7fcb76ec2a30, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=24, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcab7fe6870, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[Ljava.nio.ByteBuffer;", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=24, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fc9ee424570, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="java.nio.MappedByteBuffer", typeObj->tp_base=0x7fcb76ec2870, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=24, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fc9ee4243b0, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="java.nio.DirectByteBuffer", typeObj->tp_base=0x7fc9ee424570, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=24, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fc9ee4241f0, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="java.nio.DirectByteBufferR", typeObj->tp_base=0x7fc9ee4243b0, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=24, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
```

Notice, all the basicsize are the same, whereas I would expect something larger for the new type. Here are all the types that *aren't* 24:

```
$ grep tp_basicsize /tmp/test.txt | grep -v "tp_basicsize=24"
JType_InitSlots: typeObj=0x7fcabc3df830, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[I", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc3dfbb0, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[S", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc3fc030, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[B", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc3fc1f0, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[J", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc3fc3b0, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[D", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc3fc570, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[C", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc3fc730, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[F", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
JType_InitSlots: typeObj=0x7fcabc42b130, Py_TYPE(typeObj)=0x7fcb76cddf00, typeObj->tp_name="[Z", typeObj->tp_base=0x7fcb76ec1370, typeObj->tp_init=0x7fcc2c288bf0, typeObj->tp_basicsize=56, &JType_Type=0x7fcc2c2a04a0, &PyType_Type=0x7fcb76cddf00, JObj_init=0x7fcc2c288bf0
```
So, at least the array types are getting the correct size.

devinrsmith:
Here's a script I've been using:

```python
import jpy
from contextlib import contextmanager

_JByteBuffer = jpy.get_type('java.nio.ByteBuffer')
_JArrowToTableConverter = jpy.get_type('io.deephaven.extensions.barrage.util.ArrowToTableConverter')

@contextmanager
def jpy_flags(flags):
    orig_flags = jpy.diag.flags
    jpy.diag.flags = flags
    try:
        yield
    finally:
        jpy.diag.flags = orig_flags

def buffer_protocol():
    return jpy.byte_buffer(b'hello world')

def j_byte_buffer():
    return _JByteBuffer.allocate(42)

def j_buffer():
    # return type is java.nio.Buffer *not* ByteBuffer
    # impl is ByteBuffer.allocateDirect(size)
    return _JArrowToTableConverter.myBuffer(43)

def j_object():
    # return type is java.lang.Object 
    # impl is new Object()
    return _JArrowToTableConverter.newObject()

def print_info(name, x):
    print(f"{name}: type(x)={type(x)}" + "\n")
    print(f"{name}: x={x}" + "\n")

def create_print_del(name, fn):
    my_obj = fn()
    with jpy_flags(jpy.diag.F_OFF):
        print_info(name, my_obj)
    del my_obj

with jpy_flags(jpy.diag.F_MEM | jpy.diag.F_TYPE):
    create_print_del('j_object', j_object)
    create_print_del('buffer_protocol', buffer_protocol)
    create_print_del('j_byte_buffer', j_byte_buffer)
    create_print_del('j_buffer', j_buffer)
```

I've added some public static methods to `ArrowToTableConverter` just because it was a place to put some java logic.

devinrsmith:
https://github.com/devinrsmith/jpy/tree/111-DirectByteBuffer-support is the branch where I've added some logging.

src/main/c/jpy_jbyte_buffer.h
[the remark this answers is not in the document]
devinrsmith: Ah, yeah - I see that follows the conventions for types. I wonder if `JPy_JByteBufferStruct` might be a be a more appropriate name? Or maybe, `JPy_JBufferProtocol`, b/c there might be other types that follow this same pattern in the future? (Newer version of java have something called `MemorySegment` that are the "future" instead of ByteBuffer.)

src/main/c/jpy_jobj.c
```diff
@@ -822,3 +841,13 @@ int JType_Check(PyObject* arg)
     return PyType_Check(arg) && JPY_IS_JTYPE(arg);
 }
 
+int JByteBuffer_Check(JPy_JType* type) {
+    while (type != NULL) {
+        if (type == JPy_JByteBuffer || strcmp(type->javaName, "java.nio.ByteBuffer") == 0) {
```
devinrsmith, line 846: What's the case where `type != JPy_JByteBuffer` and `strcmp(type->javaName, "java.nio.ByteBuffer") == 0`?

src/main/c/jpy_jobj.c
```diff
@@ -727,7 +740,13 @@ int JType_InitSlots(JPy_JType* type)
...
     //Py_SET_SIZE(type, sizeof (JPy_JType));
 
-    typeObj->tp_basicsize = isPrimitiveArray ? sizeof (JPy_JArray) : sizeof (JPy_JObj);
+    if (isPrimitiveArray) {
+        typeObj->tp_basicsize = sizeof(JPy_JArray);
+    } else if (JByteBuffer_Check(type)) {
+        typeObj->tp_basicsize = sizeof(JPy_JByteBufferWrapper);
+    } else {
+        typeObj->tp_basicsize = sizeof(JPy_JObj);
+    }
```
devinrsmith, line 749: This strategy is _ok_ for now. If you feel so inclined, we _could_ also consider adding a new field `int structId` to `JPy_JType` to make this a bit more general - forcing the logic to happen in the the type-creation phase (ensuring there isn't any ambiguity and potentially mis-relying on the actual object value):

```c
int JObj_StructCheck(JPy_JType* type) {
    return type->structId == 0;
}

int JArray_StructCheck(JPy_JType* type) {
    return type->structId == 1;
}

int JByteBufferWrapper_StructCheck(JPy_JType* type) {
    return type->structId == 2;
}
```


```c
    if (JObj_StructCheck(type)) {
        typeObj->tp_basicsize = sizeof(JPy_JObj);
    } else if (JArray_StructCheck(type)) {
        typeObj->tp_basicsize = sizeof(JPy_JArray);
    } else if (JByteBufferWrapper_StructCheck(type)) {
        typeObj->tp_basicsize = sizeof(JPy_JByteBufferWrapper);
    }
```

devinrsmith (commented):
This resolves my memory safety / allocation / deallocation concerns (at least as far as the c layer itself is concerned - there are still memory safety concerns around how users _must_ hold `jpy.byte_buffer` alive as long as the ByteBuffer is in java). I think the big take away is that this struct logic _needs_ to be based on the `type` and not the actual object - it's a nuanced point that could have led to very-hard-to-debug crashes.

I want to make sure we get the naming correct; and also see if Jianfeng wants to move this to this logic to type-creation time (instead of object-creation time) which would help solidify the correct pattern. Neither are blockers for approval, but just want to poll / discuss.

src/main/c/jpy_jobj.c
jmao-denver: The `strcmp` case is really for being in the middle of building the type for `java.nio.ByteBuffer`, i.e. `JPy_JByteBuffer` isn't fully formed yet.

src/main/c/jpy_jbyte_buffer.h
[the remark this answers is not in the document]
jmao-denver: I still like `JPy_JByteBufferWrapper` a little better, but what about `JPy_JByteBufferObj`, since it is like a sub-class of `JObj`?

src/main/c/jpy_jobj.c
jmao-denver: I think it is a good idea that can wait till we have one more special type to deal with.

src/main/c/jpy_jbyte_buffer.h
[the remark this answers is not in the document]
devinrsmith: Yes - I think `JPy_JByteBufferObj` is better than the Wrapper version.

src/main/c/jpy_module.c
```diff
@@ -38,6 +39,8 @@ PyObject* JPy_destroy_jvm(PyObject* self, PyObject* args);
 PyObject* JPy_get_type(PyObject* self, PyObject* args, PyObject* kwds);
 PyObject* JPy_cast(PyObject* self, PyObject* args);
 PyObject* JPy_array(PyObject* self, PyObject* args);
+PyObject* JType_CreateJavaByteBufferObj(JNIEnv* jenv, PyObject* pyObj);
```
devinrsmith, line 42: I don't think we want this stub here - these are just the ones that need to be defined for `jpy.has_jvm()`, `jpy.array(...)` etc. In fact, I don't think it needs a stub anywhere, right?

devinrsmith (approved):

PR #128, opened 2024-01-10
Function for converting Python values to an explicit Java type

rbasralian:
- Fix a bug in `cast`
- Add an `as_jobj` method to convert a Python value to a Java object of the given type. This leverages the existing `JPy_AsJObjectWithType` method.
- Update some tests to match current behavior (e.g. returning `Byte` instead of `Integer`)
- More logging during tests
- Fixed failing `hashNegativeOne` test
- Replace incorrect `return Py_NONE`s with `Py_RETURN_NONE` (for correct reference counting)

### DIFF
--- a/jpyutil.py
+++ b/jpyutil.py
@@ -576,7 +576,6 @@ def _execute_python_scripts(scripts, **kwargs):
 
     failures = 0
     for script in scripts:
-        print('Running script ' + str(script))
         exit_code = subprocess.call([sys.executable, script], **kwargs)
         if exit_code:
             failures += 1
--- a/jpyutil.py
+++ b/jpyutil.py
@@ -36,7 +36,7 @@ import subprocess
 __author__ = "Norman Fomferra (Brockmann Consult GmbH) and contributors"
 __copyright__ = "Copyright 2015-2018 Brockmann Consult GmbH and contributors"
 __license__ = "Apache 2.0"
-__version__ = "0.15.0.dev0"
+__version__ = "0.15.0.dev1"
 
 
 # Setup a dedicated logger for jpyutil.
--- a/jpyutil.py
+++ b/jpyutil.py
@@ -576,6 +576,7 @@ def _execute_python_scripts(scripts, **kwargs):
 
     failures = 0
     for script in scripts:
+        print('Running script ' + str(script))
         exit_code = subprocess.call([sys.executable, script], **kwargs)
         if exit_code:
             failures += 1
--- a/src/main/c/jpy_jtype.c
+++ b/src/main/c/jpy_jtype.c
@@ -1447,10 +1447,10 @@ PyObject* JType_GetOverloadedMethod(JNIEnv* jenv, JPy_JType* type, PyObject* met
             } else if (type != JPy_JObject && JPy_JObject != NULL) {
                 return JType_GetOverloadedMethod(jenv, JPy_JObject, methodName, JNI_FALSE);
             } else {
-                return Py_None;
+                Py_RETURN_NONE;
             }
         } else {
-            return Py_None;
+            Py_RETURN_NONE;
         }
     }
 
--- a/src/main/c/jpy_jtype.c
+++ b/src/main/c/jpy_jtype.c
@@ -1298,6 +1298,7 @@ int JType_AddClassAttribute(JNIEnv* jenv, JPy_JType* declaringClass)
             return -1;
         }
         PyDict_SetItem(typeDict, Py_BuildValue("s", "jclass"), (PyObject*) JObj_FromType(jenv, JPy_JClass, declaringClass->classRef));
+        PyDict_SetItem(typeDict, Py_BuildValue("s", "jclassname"), (PyObject*) JPy_FROM_CSTR(declaringClass->typeObj.tp_name));
     }
     return 0;
 }
--- a/src/main/c/jpy_jtype.c
+++ b/src/main/c/jpy_jtype.c
@@ -914,19 +914,19 @@ int JType_ConvertPythonToJavaObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyA
         return JType_CreateJavaDoubleObject(jenv, type, pyArg, objectRef);
     } else if (type == JPy_JPyObject) {
         return JType_CreateJavaPyObject(jenv, type, pyArg, objectRef);
-    } else if (JPy_IS_STR(pyArg) && (type == JPy_JString || type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_JString->classRef, type->classRef)))) {
+    } else if (JPy_IS_STR(pyArg) && (type == JPy_JString || type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_JString->classRef, type->classRef))) {
         return JPy_AsJString(jenv, pyArg, objectRef);
-    } else if (PyBool_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Boolean_JClass, type->classRef)))) {
+    } else if (PyBool_Check(pyArg) && (type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_Boolean_JClass, type->classRef))) {
         return JType_CreateJavaBooleanObject(jenv, type, pyArg, objectRef);
     } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_Number_JClass, type->classRef))) {
         return JType_CreateJavaNumberFromPythonInt(jenv, type, pyArg, objectRef);
     } else if (JPy_IS_CLONG(pyArg) && ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef))) {
         return JType_CreateJavaIntegerObject(jenv, type, pyArg, objectRef);
-    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)))) {
+    } else if (JPy_IS_CLONG(pyArg) && (*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)) {
         return JType_CreateJavaLongObject(jenv, type, pyArg, objectRef);
-    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Double_JClass, type->classRef)))) {
+    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_Double_JClass, type->classRef))) {
         return JType_CreateJavaDoubleObject(jenv, type, pyArg, objectRef);
-    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Float_JClass, type->classRef)))) {
+    } else if (PyFloat_Check(pyArg) && (*jenv)->IsAssignableFrom(jenv, JPy_Float_JClass, type->classRef)) {
         return JType_CreateJavaFloatObject(jenv, type, pyArg, objectRef);
     } else if (type == JPy_JObject && allowObjectWrapping) {
         return JType_CreateJavaPyObject(jenv, JPy_JPyObject, pyArg, objectRef);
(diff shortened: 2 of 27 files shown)


jpyutil.py
```diff
@@ -36,7 +36,7 @@
 __author__ = "Norman Fomferra (Brockmann Consult GmbH) and contributors"
 __copyright__ = "Copyright 2015-2018 Brockmann Consult GmbH and contributors"
 __license__ = "Apache 2.0"
-__version__ = "0.15.0"
+__version__ = "0.16.0.dev0"
```
devinrsmith, line 39: Merge the latest master, this should be taken care of.

src/main/c/jpy_module.c
```diff
@@ -594,6 +624,76 @@ PyObject* JPy_cast(PyObject* self, PyObject* args)
...
+    if (objectRef == NULL) {
+        PyErr_NoMemory();
+        return NULL;
+    }
+
+    // Create a PyObject (JObj) to hold the result
+    resultObj = (JPy_JObj*) PyObject_New(JPy_JObj, JTYPE_AS_PYTYPE(targetTypeParsed));
+    if (resultObj == NULL) {
+        return NULL;
+    }
```
devinrsmith, line 684: We should delete the global ref if this happens.

src/main/c/jpy_module.c
```diff
@@ -57,10 +59,19 @@ static PyMethodDef JPy_Functions[] = {
...
 
+    {"get_type_name",    (PyCFunction) JPy_get_type_name, METH_VARARGS,
+                    "get_type_name(obj) - Return the Java type associated with the 'obj' reference. "
+                    "This is reference's declared type (which may be a supertype), rather than the object's runtime type."},
+
     {"cast",        JPy_cast, METH_VARARGS,
                     "cast(obj, type) - Cast the given Java object to the given Java type (type name or type object). "
                     "Returns None if the cast is not possible."},
 
+    {"as_jobj",     JPy_as_jobj, METH_VARARGS,
```
devinrsmith, line 70: This is cool - previously, we _could_ do this, it just required explicit java code:

```java
public class MyClass {
  public static MyJavaType identity(MyJavaType x) {
    return x;
  }
}
```

```python
jobj = jpy.get_type('MyClass').identity(pobj)
```

I'm assuming these are the semantics you are after?

src/test/python/jpy_typeconv_test.py
```diff
@@ -16,14 +16,58 @@ def setUp(self):
...
+        my_jcharseq1 = jpy.cast(my_jstr, 'java.lang.CharSequence')
+        self.assertEqual(jpy.get_type_name(my_jcharseq1), 'java.lang.CharSequence')
+        self.assertEqual(fixture.stringifyObjectArg(my_jcharseq1), 'String(testStr)')
+
+        my_jcharseq2 = jpy.cast(my_jstr, jpy.get_type('java.lang.CharSequence'))
+        self.assertEqual(jpy.get_type_name(my_jcharseq2), 'java.lang.CharSequence')
+        self.assertEqual(fixture.stringifyObjectArg(my_jcharseq2), 'String(testStr)')
+
+
+    def test_ToObjectConversionTyped(self):
```
devinrsmith, line 43: I think trying to exercise some of the other logic from JType_ConvertPythonToJavaObject would be nice; arrays, boolean, char, PyObject, String.

src/main/c/jpy_module.c
```diff
@@ -57,10 +59,19 @@ static PyMethodDef JPy_Functions[] = {
                     "get_type(name, resolve=True) - Return the Java class with the given name, e.g. 'java.io.File'. "
                     "Loads the Java class from the JVM if not already done. Optionally avoids resolving the class' methods."},
 
+    {"get_type_name",    (PyCFunction) JPy_get_type_name, METH_VARARGS,
```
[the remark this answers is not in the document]
devinrsmith: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_name

Looks like
```python
>>> f"{type.__module__}.{type.__name}"
```

is the way to get the java class name; this doesn't quite work for primitives (not sure what we expect to happen w/ primitives).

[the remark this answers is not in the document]
rbasralian: I don't think primitives apply — I don't think you can wrap an unboxed primitive value in a `JPy_JObj`.

I got rid of this method and added `type(obj).jclassname`.

src/main/c/jpy_module.c
rbasralian: Thanks — fixed

src/main/c/jpy_module.c
rbasralian: Yeah, basically I had two parallel lists, one of java types and one of python values (in practice strings/numbers), and I wanted to stick the python values into a java `Object[]` with the given Java type, but numbers I wanted as `Long` were getting converted to `Byte`. I could have just used the constructors explicitly but that's not as clean, and it seems worthwhile to expose `JType_ConvertPythonToJavaObject` anyway.

src/test/python/jpy_typeconv_test.py
rbasralian: Added more tests. The test for converting explicitly to `PyObject` doesn't run, apparently due to an issue with the classpath

src/test/java/org/jpy/PyObjectTest.java
```diff
@@ -628,7 +632,23 @@ public void hashNegativeOne() {
...
+
+
+        /*
+        java.lang.RuntimeException: Error in Python interpreter:
+        Type: <class 'ValueError'>
+        Value: cannot convert a Python 'HashNegativeOne' to a Java 'java.lang.Object'
+         */
+//        System.out.println(obj.getObjectValue());
+
+//        assertEquals(-1, obj.getIntValue()); // TODO: should this be 'result.getIntValue()'?
```
devinrsmith, line 651: Yes; this should be result.getIntValue()

src/test/python/jpy_typeconv_test.py
```diff
@@ -3,27 +3,338 @@
...
+        self.assertEqual(fixture.stringifyObjectArg(jobj), 'Float[](Float(12.0),Float(13.0),Float(14.0))')
+
+        target_type = type(jpy.array(jpy.get_type('java.lang.Double'), 0))
+        jobj = jpy.as_jobj([12, 13, 14], target_type)
+        self.assertEqual(type(jobj), target_type)
+        self.assertEqual(fixture.stringifyObjectArg(jobj), 'Double[](Double(12.0),Double(13.0),Double(14.0))')
+
+    def test_AsJobjToPyObject(self):
+        print('Starting test_AsJobjToPyObject')
+        # TODO: "Java class 'org.jpy.PyObject' not found"???
```
devinrsmith, line 214: This makes sense; PyObject is only necessary if the user needs to reference python from java. At least with this test scaffolding, the jpy jar is not on the classpath.

devinrsmith (commented):
I don't think your changes to the JVOID / JNULL macros match with their indented use cases. While technically the results will be the same, the usage semantics are wrong. I would revert all these changes.

As you've probably noticed, the jpy tests aren't very robust. Ideally, we'd get them to run in GH CI here, but that hasn't been done yet.

src/main/c/jpy_jtype.c
```diff
@@ -914,19 +914,19 @@ int JType_ConvertPythonToJavaObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyA
...
     } else if (JPy_IS_CLONG(pyArg) && ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef))) {
         return JType_CreateJavaIntegerObject(jenv, type, pyArg, objectRef);
-    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)))) {
+    } else if (JPy_IS_CLONG(pyArg) && (*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)) {
         return JType_CreateJavaLongObject(jenv, type, pyArg, objectRef);
-    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Double_JClass, type->classRef)))) {
+    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_Double_JClass, type->classRef))) {
         return JType_CreateJavaDoubleObject(jenv, type, pyArg, objectRef);
-    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Float_JClass, type->classRef)))) {
+    } else if (PyFloat_Check(pyArg) && (*jenv)->IsAssignableFrom(jenv, JPy_Float_JClass, type->classRef)) {
```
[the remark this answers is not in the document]
rbasralian: the cases I changed were:

```
    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_Number_JClass, type->classRef))) {
        return JType_CreateJavaNumberFromPythonInt(jenv, type, pyArg, objectRef);
    } else if (JPy_IS_CLONG(pyArg) && ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef))) {
        return JType_CreateJavaIntegerObject(jenv, type, pyArg, objectRef);
    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)))) {
        return JType_CreateJavaLongObject(jenv, type, pyArg, objectRef);
```

and:

```
    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Double_JClass, type->classRef)))) {
        return JType_CreateJavaDoubleObject(jenv, type, pyArg, objectRef);
    } else if (PyFloat_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Float_JClass, type->classRef)))) {
        return JType_CreateJavaFloatObject(jenv, type, pyArg, objectRef);
    }
```

in both case, the second `type == JPy_JObject` test will always be false — for the first case, we will have already landed in `JType_CreateJavaNumberFromPythonInt` and in the second case we'd hit `JType_CreateJavaDoubleObject`.

src/test/python/jpy_typeconv_test.py
rbasralian: Ok — I moved this to a separate file (`jpy_typeconv_test_pyobj.py`).

src/test/java/org/jpy/PyObjectTest.java
rbasralian: Thanks — updated & reenabled the test

src/main/c/jpy_module.c
```diff
@@ -451,7 +457,7 @@ PyObject* JPy_create_jvm(PyObject* self, PyObject* args, PyObject* kwds)
     if (JPy_JVM != NULL) {
         JPy_DIAG_PRINT(JPy_DIAG_F_JVM + JPy_DIAG_F_ERR, "JPy_create_jvm: WARNING: Java VM is already running.\n");
         JPy_DECREF(options);
-        return Py_BuildValue("");
+        return JPy_FROM_JVOID();
```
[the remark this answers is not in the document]
rbasralian: Replaced the handful of `Py_BuildValue()`s I changed with a new `JPy_PY_NONE()`

devinrsmith (commented):
LGTM. Adding Jianfeng for final pass.

jmao-denver (commented):
The changes LGTM, not sure if I like the name very much.

src/main/c/jpy_module.c
rbasralian: I'm definitely open to changing it. I went with `as_jobj` because it ultimately calls `JPy_AsJObjectWithType`, and I thought having 'obj' in the name explicitly could be helpful because [conversion](https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html) in Java is a distinct concept (though a related one) that also involves Java primitives.

@devinrsmith what do you think?

devinrsmith: I think "as" gives off the wrong semantic notion, somewhat akin to zero-cost. While the same could be said about `JPy_AsJObjectWithType`, that's lower level and not exposed to the end user. Also, it calls down into `JType_ConvertPythonToJavaObject`.

I'm also not the biggest fan of trying to save a few characters; my vote would be for `jpy.convert` instead of `jpy.conv` / `jpy.as_jobj`.

devinrsmith (approved):
LGTM.

jmao-denver (approved):
I am happy with the name change.

PR #145, opened 2024-04-30
Use Py_ssize_t when calculate buffer len

### LINKED ISSUE #143: Large arrays create incorrect buffers
This issue was originally noted in https://github.com/deephaven/deephaven-core/issues/5403.  There are many details there.

Reproducer:
```python

import jpy
import numpy as np

i_max = 2_147_483_647

repeat = jpy.get_type("io.deephaven.function.Basic").repeat

def test_it(name, size, jtype, np_type):
    print(f"TEST: {name} || {jtype} {np_type} {size}")

    narray = np.ones(size, dtype=np_type)
    jarray = jpy.array(jtype, narray)

    try:
        buf = narray.tobytes()
        np.frombuffer(buf, dtype=np_type)
        print("\tPASS - np buffer")
    except Exception as e:
        print("\tFAIL - np buffer")
        print(f"\t\tERROR: {e}")

    try:
        np.frombuffer(jarray, dtype=np_type)
        print("\tPASS - jpy buffer")
    except Exception as e:
        print("\tFAIL - jpy buffer")
        print(f"\t\tERROR: {e}")

test_it("Float64 small", i_max//8, "double", np.float64)
test_it("Float64 large", i_max//8 + 1, "double", np.float64)

test_it("Int64 small", i_max//8, "long", np.int64)
test_it("Int64 large", i_max//8 + 1, "long", np.int64)
```

Output:
```
TEST: Float64 small || double <class 'numpy.float64'> 268435455
	PASS - np buffer
	PASS - jpy buffer
TEST: Float64 large || double <class 'numpy.float64'> 268435456
	PASS - np buffer
	FAIL - jpy buffer
		ERROR: offset must be non-negative and no greater than buffer length (-2147483648)
TEST: Int64 small || long <class 'numpy.int64'> 268435455
	PASS - np buffer
	PASS - jpy buffer
TEST: Int64 large || long <class 'numpy.int64'> 268435456
	PASS - np buffer
	FAIL - jpy buffer
		ERROR: offset must be non-negative and no greater than buffer length (-2147483648)
```

The buffer creation yields incorrect array lengths when arrays are longer than `MAX_INT/TYPE_SIZE_IN_BYTES`.  Most likely a type that is too small is being used to compute lengths.

Note that numpy created arrays of the same size appear to function properly.

jmao-denver:
Also fixed all the pointer to long conversions that clang 15 considers error by default on MacOS M2

Fixes #143

### DIFF
--- a/src/main/c/jni/org_jpy_PyLib.c
+++ b/src/main/c/jni/org_jpy_PyLib.c
@@ -1021,7 +1021,7 @@ jlong JNICALL Java_org_jpy_PyLib_executeCode
     codeChars = (*jenv)->GetStringUTFChars(jenv, jCode, NULL);
     if (codeChars == NULL) {
         PyLib_ThrowOOM(jenv);
-        return NULL;
+        return 0;
     }
 
     JPy_DIAG_PRINT(JPy_DIAG_F_EXEC, "Java_org_jpy_PyLib_executeCode: code='%s'\n", codeChars);
--- a/src/main/c/jpy_diag.c
+++ b/src/main/c/jpy_diag.c
@@ -103,7 +103,7 @@ PyTypeObject Diag_Type =
     sizeof (JPy_Diag),            /* tp_basicsize */
     0,                            /* tp_itemsize */
     NULL,                         /* tp_dealloc */
-    NULL,                         /* tp_print */
+    0,                            /* tp_print */
     NULL,                         /* tp_getattr */
     NULL,                         /* tp_setattr */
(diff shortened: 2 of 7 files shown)


devinrsmith (approved):
I think there's an issue associated with this PR?

jmao-denver:
> I think there's an issue associated with this PR?

Yes, 5403, I manually tested it and confirmed the fix. But I had to increase the max JVM heap size to 8G, I am not sure if we want to include the test in the CI.

PR #146, opened 2024-05-01
Delay resolving super classes until referenced 

### LINKED ISSUE #142: get_type ordering crash
In some cases, it seems like `jpy.get_type` will cause crashes and SIGSEGVs. During the development of https://github.com/deephaven/deephaven-core/pull/5225, I noticed that the following crashes:

```python
import jpy

jpy.get_type("io.deephaven.json.ArrayOptions")
```

, but works if we import the base class first:

```python
import jpy

jpy.get_type("io.deephaven.json.ValueOptions")
jpy.get_type("io.deephaven.json.ArrayOptions")
```

The error looks something like

```
2024-04-10T22:59:04.700Z | r-Scheduler-Serial-1 |  INFO | .p.PythonDeephavenSession | Evaluating command: import jpy
2024-04-10T22:59:06.225Z | r-Scheduler-Serial-1 |  INFO | .p.PythonDeephavenSession | Evaluating command: jpy.get_type("io.deephaven.json.ArrayOptions")
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fc317a0f9e9, pid=85835, tid=86053
#
# JRE version: OpenJDK Runtime Environment Zulu21.32+17-CA (21.0.2+13) (build 21.0.2+13-LTS)
# Java VM: OpenJDK 64-Bit Server VM Zulu21.32+17-CA (21.0.2+13-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# C  [libpython3.12.so+0x20f9e9]
...
```

I assumed there might have been an issues with some static initializers, but I couldn't find anything. Via groovy, importing ArrayOptions first seems to work out just fine.

jmao-denver:
Fixes #142

1. be able to generate Python types from Java classes with complicated cyclic references
2. fix a reference counting issue when a new Python type fails to finalize
3. fix the Python tests and disable the Java tests due to previous changes (years ago) to enable auto PyObjects cleanup in Java.
4. add a limited GH check workflow to run the Python tests on Linux
5. add two new Python test cases

### DIFF
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -726,11 +726,13 @@ int JType_InitSlots(JPy_JType* type)
 
     typeObj = JTYPE_AS_PYTYPE(type);
 
-    #if defined(JPY_COMPAT_39P)
-        Py_SET_REFCNT(typeObj, 1);
-    #else
-        Py_REFCNT(typeObj) = 1;
-    #endif
+    // This is hacky to make the caller JType_GetType be able to call JPy_INCREF regardless if the type was already created,
+    // but it will cause a ref counting error when this function fails, and we need to remove the type from the registry
+    //    #if defined(JPY_COMPAT_39P)
+    //        Py_SET_REFCNT(typeObj, 1);
+    //    #else
+    //        Py_REFCNT(typeObj) = 1;
+    //    #endif
     Py_SET_TYPE(typeObj, NULL);
     Py_SET_SIZE(typeObj, 0);
     // todo: The following lines are actually correct, but setting Py_TYPE(type) = &JType_Type results in an interpreter crash. Why?
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -726,14 +726,6 @@ int JType_InitSlots(JPy_JType* type)
 
     typeObj = JTYPE_AS_PYTYPE(type);
 
-    // This is hacky to make the caller JType_GetType able to call JPy_INCREF regardless if the type was already created,
-    // but it will cause a ref counting error when this function fails, and we need to remove the type from the registry.
-    // The proper ref counting needs to be done in JType_GetType().
-    //    #if defined(JPY_COMPAT_39P)
-    //        Py_SET_REFCNT(typeObj, 1);
-    //    #else
-    //        Py_REFCNT(typeObj) = 1;
-    //    #endif
     Py_SET_TYPE(typeObj, NULL);
     Py_SET_SIZE(typeObj, 0);
     // todo: The following lines are actually correct, but setting Py_TYPE(type) = &JType_Type results in an interpreter crash. Why?
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -726,7 +726,7 @@ int JType_InitSlots(JPy_JType* type)
 
     typeObj = JTYPE_AS_PYTYPE(type);
 
-    // This is hacky to make the caller JType_GetType be able to call JPy_INCREF regardless if the type was already created,
+    // This is hacky to make the caller JType_GetType able to call JPy_INCREF regardless if the type was already created,
     // but it will cause a ref counting error when this function fails, and we need to remove the type from the registry
     //    #if defined(JPY_COMPAT_39P)
     //        Py_SET_REFCNT(typeObj, 1);
--- a/src/main/c/jpy_jobj.c
+++ b/src/main/c/jpy_jobj.c
@@ -727,7 +727,8 @@ int JType_InitSlots(JPy_JType* type)
     typeObj = JTYPE_AS_PYTYPE(type);
 
     // This is hacky to make the caller JType_GetType able to call JPy_INCREF regardless if the type was already created,
-    // but it will cause a ref counting error when this function fails, and we need to remove the type from the registry
+    // but it will cause a ref counting error when this function fails, and we need to remove the type from the registry.
+    // The proper ref counting needs to be done in JType_GetType().
     //    #if defined(JPY_COMPAT_39P)
     //        Py_SET_REFCNT(typeObj, 1);
     //    #else
--- a/src/main/c/jpy_jtype.c
+++ b/src/main/c/jpy_jtype.c
@@ -171,7 +171,7 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
         //printf("T2: type->tp_init=%p\n", ((PyTypeObject*)type)->tp_init);
 
         // ... before we can continue processing the super type ...
-        if (JType_InitSuperType(jenv, type, resolve) < 0) {
+        if (JType_InitSuperType(jenv, type, JNI_FALSE) < 0) {
             PyDict_DelItem(JPy_Types, typeKey);
             return NULL;
         }
--- a/src/main/c/jpy_jtype.c
+++ b/src/main/c/jpy_jtype.c
@@ -156,7 +156,7 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
 
         found = JNI_FALSE;
 
-        // Create a new type instance
+        // Create a new type instance, the refcount is set to 1 if this call succeeds
         type = JType_New(jenv, classRef, resolve);
         if (type == NULL) {
             JPy_DECREF(typeKey);
@@ -166,6 +166,7 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
         //printf("T1: type->tp_init=%p\n", ((PyTypeObject*)type)->tp_init);
 
         // In order to avoid infinite recursion, we have to register the new (but yet incomplete) type first...
+        // it will increment the refcount of type
         PyDict_SetItem(JPy_Types, typeKey, (PyObject*) type);
 
         //printf("T2: type->tp_init=%p\n", ((PyTypeObject*)type)->tp_init);
@@ -209,8 +210,9 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
         }
 
         JType_AddClassAttribute(jenv, type);
-        // 'type' will be returned as a new reference. 'typeKey' needs to be released.
+        // 'type' will be refcount incremented and returned as a new reference later. 'typeKey' needs to be released.
         JPy_DECREF(typeKey);
+        JPy_DECREF(type);
 
         //printf("T5: type->tp_init=%p\n", ((PyTypeObject*)type)->tp_init);
 
@@ -233,8 +235,7 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
         }
 
         type = (JPy_JType*) typeValue;
-        // 'type' will be returned as a new reference. 'typeKey' needs to be released.
-        JPy_INCREF(type);
+        // 'type' will be refcount incremented and returned as a new reference. 'typeKey' needs to be released.
         JPy_DECREF(typeKey);
     }
 
@@ -242,12 +243,12 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
 
     if (!type->isResolved && resolve) {
         if (JType_ResolveType(jenv, type) < 0) {
-            JPy_DECREF(type);
             return NULL;
         }
     }
 
-     return type;
+    JPy_INCREF(type);
+    return type;
 }
 
 /**
@@ -1093,7 +1094,6 @@ int JType_InitComponentType(JNIEnv* jenv, JPy_JType* type, jboolean resolve)
         if (type->componentType == NULL) {
             return -1;
         }
-        JPy_INCREF(type->componentType);
     } else {
         type->componentType = NULL;
     }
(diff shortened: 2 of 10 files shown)


cpwright:
We should attempt to add a unit test for this.  If we understand have a deep enough understanding of the underlying problem, I expect we could have simple fixture classes to demonstrate it.

niloc132:
Worth noting that we don't run jpy tests at this time as part of CI or releases, they don't consistently pass enough to be useful. Deephaven-core has its own modified suite that we run instead.

src/main/c/jpy_jobj.c
```diff
@@ -726,11 +726,13 @@ int JType_InitSlots(JPy_JType* type)
 
     typeObj = JTYPE_AS_PYTYPE(type);
 
-    #if defined(JPY_COMPAT_39P)
-        Py_SET_REFCNT(typeObj, 1);
-    #else
-        Py_REFCNT(typeObj) = 1;
-    #endif
+    // This is hacky to make the caller JType_GetType be able to call JPy_INCREF regardless if the type was already created,
```
chipkent, line 729: 1. Do you want to leave this commented out stuff in here?
2. I'm not sure it is hacky after reading the code.

src/test/java/org/jpy/fixtures/CyclicReferenceParent.java
```diff
@@ -0,0 +1,32 @@
...
+// Copyright (c) 2016-2024 Deephaven Data Labs and Patent Pending
+//
+
+package org.jpy.fixtures;
+
+import java.lang.reflect.Array;
+import java.lang.reflect.Method;
+
+/**
+ * Used as a test class for the test cases in jpy_overload_test.py
```
devinrsmith, line 11: wrong python file?

src/test/java/org/jpy/fixtures/CyclicReferenceGrandParent.java
```diff
@@ -0,0 +1,33 @@
...
+// Copyright (c) 2016-2024 Deephaven Data Labs and Patent Pending
+//
+
+package org.jpy.fixtures;
+
+import java.lang.reflect.Array;
+import java.lang.reflect.Method;
+
+/**
+ * Used as a test class for the test cases in jpy_overload_test.py
```
devinrsmith, line 11: wrong python file?

src/test/java/org/jpy/fixtures/CyclicReferenceChild2.java
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) 2016-2024 Deephaven Data Labs and Patent Pending
+//
+
+package org.jpy.fixtures;
+
+import java.lang.reflect.Array;
+import java.lang.reflect.Method;
+
+/**
+ * Used as a test class for the test cases in jpy_overload_test.py
```
devinrsmith, line 11: wrong python file?

src/test/java/org/jpy/fixtures/CyclicReferenceChild1.java
```diff
@@ -0,0 +1,30 @@
...
+// Copyright (c) 2016-2024 Deephaven Data Labs and Patent Pending
+//
+
+package org.jpy.fixtures;
+
+import java.lang.reflect.Array;
+import java.lang.reflect.Method;
+
+/**
+ * Used as a test class for the test cases in jpy_overload_test.py
```
devinrsmith, line 11: wrong python file?

src/test/python/jpy_array_test.py
```diff
@@ -232,6 +232,12 @@ def test_leak(self):
         for i in range(1000000):
             memory_view = memoryview(j_int_array)
 
+    def test_size_greater_than_maxint(self):
+        jarr = jpy.array("int", 2**30)
+        mv = memoryview(jarr)
+        self.assertEqual(mv.nbytes, 2**32)
```
devinrsmith, line 238: This change seems unrelated to the issue?

src/main/c/jpy_jobj.c
devinrsmith: I want to dig into this further before we merge.

.github/workflows/check.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '8'
```
devinrsmith, line 20: Probably makes sense to do matrix w/ python and java versions.

.github/workflows/check.yml
devinrsmith, line 1: This file seems unrelated to the issue at hand? We might want to pull it out as a separate PR and merge first?

setup.py
```diff
@@ -277,7 +281,8 @@ def test_java(self):
 
     suite.addTest(test_python_with_java_runtime)
     suite.addTest(test_python_with_java_classes)
-    suite.addTest(test_java)
+    # comment out because the asynchronous nature of the PyObject GC in Java makes stopPython/startPython flakey.
+    # suite.addTest(test_java)
```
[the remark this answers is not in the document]
jmao-denver: I did, no luck. May need a different test strategy.

src/main/c/jpy_jobj.c
```diff
@@ -726,11 +726,14 @@ int JType_InitSlots(JPy_JType* type)
...
-        Py_REFCNT(typeObj) = 1;
-    #endif
+    // This is hacky to make the caller JType_GetType able to call JPy_INCREF regardless if the type was already created,
+    // but it will cause a ref counting error when this function fails, and we need to remove the type from the registry.
+    // The proper ref counting needs to be done in JType_GetType().
+    //    #if defined(JPY_COMPAT_39P)
+    //        Py_SET_REFCNT(typeObj, 1);
+    //    #else
+    //        Py_REFCNT(typeObj) = 1;
+    //    #endif
```
chipkent, line 736: Does this commented out code need to stay?

src/main/c/jpy_jtype.c
```diff
@@ -215,20 +232,22 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
...
+        // 'type' will be returned as a new reference. 'typeKey' needs to be released.
+        JPy_INCREF(type);
+        JPy_DECREF(typeKey);
     }
 
     JPy_DIAG_PRINT(JPy_DIAG_F_TYPE, "JType_GetType: javaName=\"%s\", found=%d, resolve=%d, resolved=%d, type=%p\n", type->javaName, found, resolve, type->isResolved, type);
 
     if (!type->isResolved && resolve) {
         if (JType_ResolveType(jenv, type) < 0) {
+            JPy_DECREF(type);
```
chipkent, line 245: Does this need an error log?

src/main/c/jpy_jobj.c
jmao-denver: Probably not once @devinrsmith can confirm.

src/main/c/jpy_jtype.c
jmao-denver: Probably not b/c the functions that resolves the members (methods, fields, etc.) log the specific errors. That said it wouldn't hurt to log it at this level.

src/main/c/jpy_jtype.c
```diff
@@ -171,16 +171,29 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
...
+            JPy_DECREF(typeKey);
+            JPy_DECREF(type);
             return NULL;
         }
 
         //printf("T3: type->tp_init=%p\n", ((PyTypeObject*)type)->tp_init);
 
         // ... and processing the component type.
-        if (JType_InitComponentType(jenv, type, resolve) < 0) {
+        if (JType_InitComponentType(jenv, type, JNI_FALSE) < 0) {
```
devinrsmith, line 192: I think the logic around resolving component types is now incorrect when `resolve=true` (I think it's always been incorrect when `resolve=false`). I think we need to ensure that `JType_ResolveType` has similar logic to recursively resolve component types just like it does for super types.

src/main/c/jpy_jobj.c
devinrsmith: I'll need to review this further; but I have other review points that take precedence IMO.

devinrsmith (commented):
At first, I thought "resolve" was a proxy to java `Class.forName` "initialize" param, but that is _not_ the case. "resolve" here is jpy-logic for "have I reflected over the class type so I can know all the methods I can call".

It _is_ user accessible via `jpy.get_type("SomeClass", resolve=...)`. We might want to add some tests around this:

```python

AClass = jpy.get_type("AClass", resolve=False)
BClass = jpy.get_type("BClass", resolve=False) # extends AClass
CClass = jpy.get_type("CClass", resolve=False) # extends BClass

# tests to show the above classes are not resolved

jpy.get_type("BClass")

# tests to show that AClass and BClass are resolved; and CClass is still not resolved

jpy.get_type("CClass")

# tests to show that AClass, BClass, CClass are resolved
```

src/main/c/jpy_jtype.c
```diff
@@ -171,16 +171,29 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
...
         // ... before we can continue processing the super type ...
-        if (JType_InitSuperType(jenv, type, resolve) < 0) {
+        // Note, at this point, the Python type has been registered-but-yet-finalized (as done in JType_InitSlots).
+        // We need to delay resolving the super classes to when they are actually referenced, so that in a cyclic
+        // reference scenario, the super classes can still be finalized (reference-able by child classes), but NOT RESOLVED
+        // (its methods/fields remain absent in the type object). This is to avoid the case where its members reference some other
+        // registered-but-yet-finalized classes in the same class hierarchy, and these classes in turn have
+        // registered-but-yet-finalized super classes, causing JType_InitSlots to fail for one of them because its super
+        // class is not finalized. When this happens, the affected classes will not be properly resolved.
+        if (JType_InitSuperType(jenv, type, JNI_FALSE) < 0) {
```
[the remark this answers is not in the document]
jmao-denver: I think leaving the comments here (the calling place with the arguments) makes more sense. I also think although we don't know if `JType_InitSuperType` will ever be called with `resolve=True` in the future, having that option isn't a bad thing.

src/main/c/jpy_jtype.c
jmao-denver: I don't think it is incorrect. If we choose to resolve the component type here, we could run into the similar situations with the resolving the super classes.

A[] -> resolve A -> A has a field of type B[] -> resolve B -> B has a field of type A[] (which is registered but not finalized yet).

devinrsmith: I'm not saying we need to resolve _here_; I _think_ we need to resolve the component type in `JType_ResolveType` (just as we do w/ base / super type).

jmao-denver: `JType_ResolveType` doesn't actually resolve anything other than the super types (because of method resolution). Other referenced types can all be resolved at the time of reference with better or no performance implication. Added a test case for component class resolution.

devinrsmith: Okay, I can see why actually resolving component types isn't important (b/c they don't have methods, constructors, fields). It also means that `JType_InitComponentType` having `resolve` is a bit weird, as resolving doesn't really do anything... we are never calling it w/ `true` anymore as it doesn't really matter.

src/main/c/jpy_jobj.c
devinrsmith: I think by the time the code gets here, the ref count is already 1. In JType_GetType, setting the type as a dictionary value will increase the ref count:

```c
PyDict_SetItem(JPy_Types, typeKey, (PyObject*) type);
```

But also in that regard, I _think_ we are decrementing the ref below 0 in `JType_GetType` on the error cases.

Currently, we do:

```c
            PyDict_DelItem(JPy_Types, typeKey);
            JPy_DECREF(typeKey);
            JPy_DECREF(type);
```

but I think we should be doing

```c
            PyDict_DelItem(JPy_Types, typeKey);
            JPy_DECREF(typeKey);
```

In essence, the `JPy_Types` dictionary is what _should_ be keeping the type alive.

I think a bit of re working of `JType_GetType` for refcounting would make this better.

But also, if we have the type in the dict, but then `JType_ResolveType` fails, should we remove it from the dict?

devinrsmith (commented):
I'm also noticing we may be overcounting in `JType_InitComponentType` and `JType_InitSuperType`:

```c
    if (superClassRef != NULL) {
        type->superType = JType_GetType(jenv, superClassRef, resolve);
        if (type->superType == NULL) {
            return -1;
        }
        JPy_INCREF(type->superType);
        JPy_DELETE_LOCAL_REF(superClassRef);
    }
```

`JType_GetType` should be incrementing the ref count for us, so we shouldn't need to increment it again, right?

src/test/python/jpy_gettype_test.py
```diff
@@ -71,6 +71,34 @@ def test_issue_74(self):
...
+        j_child1 = j_child1_class.of(8)
+        j_child2s = j_child1.getChild2s()
+        self.assertIn("[Lorg.jpy.fixtures.CyclicReferenceChild2;", repr(type(j_child2s)))
+        for j_child2 in j_child2s:
+            self.assertTrue(j_child2.getName().startswith("Child2"))
+
+    def test_fail_init_supertype(self):
+        with self.assertRaises(ValueError) as cm:
+            j_child_class = jpy.get_type("org.jpy.fixtures.GetTypeFailureChild")
+        print(str(cm.exception))
```
jmao-denver, line 101:
```suggestion
```

chipkent (approved):

PR #158, opened 2024-08-08
Get the correct computed tb lineno

jmao-denver:
Fix for https://github.com/python/cpython/issues/109181 introduced lazily computed lineno for traceback object in 3.11.7 and 3.12.1. Tested in a number of Python versions, the change seems to be safe.

The problem was initially reported in https://github.com/deephaven/vscode-deephaven/issues/9

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,7 +25,7 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
 
-      - run: pip install setuptools
+      - run: pip install "setuptools < 72"
 
       - name: Run Test
         run: python setup.py test
--- a/src/main/c/jni/org_jpy_PyLib.c
+++ b/src/main/c/jni/org_jpy_PyLib.c
@@ -2759,9 +2759,10 @@ static int format_python_traceback(PyTracebackObject *tb, char **buf, int *bufLe
     }
     while (tb != NULL && err == 0) {
         PyCodeObject* co = PyFrame_GetCode(tb->tb_frame);
+        int tb_lineno = get_traceback_lineno(tb);
         if (last_file == NULL ||
             co->co_filename != last_file ||
-            last_line == -1 || get_traceback_lineno(tb) != last_line ||
+            last_line == -1 || tb_lineno != last_line ||
             last_name == NULL || co->co_name != last_name) {
             if (cnt >  PYLIB_RECURSIVE_CUTOFF) {
                 pyObjUtf8 = format_line_repeated(cnt);
@@ -2772,7 +2773,7 @@ static int format_python_traceback(PyTracebackObject *tb, char **buf, int *bufLe
                 }
             }
             last_file = co->co_filename;
-            last_line = get_traceback_lineno(tb);
+            last_line = tb_lineno;
             last_name = co->co_name;
             cnt = 0;
         }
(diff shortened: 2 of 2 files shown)


src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -2754,7 +2761,7 @@ static int format_python_traceback(PyTracebackObject *tb, char **buf, int *bufLe
         PyCodeObject* co = PyFrame_GetCode(tb->tb_frame);
         if (last_file == NULL ||
             co->co_filename != last_file ||
-            last_line == -1 || tb->tb_lineno != last_line ||
+            last_line == -1 || get_traceback_lineno(tb) != last_line ||
```
chipkent, line 2764: The attribute get plus conversion is happening 3x.  Should we just do it once and keep the line number?

chipkent (approved):

.github/workflows/check.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
 
-      - run: pip install setuptools
+      - run: pip install "setuptools < 72"
```
devinrsmith, line 28: Why do we need to limit setuptools?

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -2730,6 +2730,13 @@ static int append_to_java_message(PyObject * pyObjUtf8, char **buf, int *bufLen
     return 0;
 }
 
+static int get_traceback_lineno(PyTracebackObject *tb) {
+    PyObject* po_lineno = PyObject_GetAttrString((PyObject*)tb, "tb_lineno");
+    int lineno = (int)PyLong_AsLong(po_lineno);
+    JPy_DECREF(po_lineno);
+    return lineno;
+}
```
devinrsmith, line 2738: Took me a little bit to figure out what actually changed. https://github.com/python/cpython/pull/111548 is a useful reference. I'm a bit surprised there's not a more official stable API around PyTracebackObject. I understand the need to wash through `PyObject_GetAttrString` in this case, but I'm sad.

Does calling `PyObject_GetAttrString` actually _set_ `tb->tb_lineno`? I can't tell from the linked PR if that's actually what python does in this case. From the description of the patch

> Speed up :obj:`Traceback` object creation by lazily compute the line number.

it's _seems_ to me that "lazily" means that it _will_ be cached after first access? (If that is _not_ the case, it seems like `tb->tb_lineno` is completely useless and unused?)

Assuming "yes", we could probably check `tb->tb_lineno == -1` before doing the more expensive call?

devinrsmith (commented):
This does point out that we need to be careful when accessing c python structs that don't have proper / stable APIs. To be technically safe, we would need to be using using `PyObject_GetAttrString` to access attributes everywhere instead of assuming simple c struct access.

For example, while we are "ok" accessing `tb->tb_next` as a proxy for the `tb_next` attribute (https://docs.python.org/3/reference/datamodel.html#traceback.tb_next) a similar change could happen in the future where we'd need to go the official route to access the attribute. https://github.com/python/cpython/blob/v3.12.5/Python/traceback.c#L100-L108

.github/workflows/check.yml
jmao-denver: See https://github.com/pypa/setuptools/issues/4519 and a solution is out of scope for now. will create a ticket.

src/main/c/jni/org_jpy_PyLib.c
jmao-denver: No, and PyTracebackObject isn't part of the public stable API, so they are free to change the struct itself and how to use it.

.github/workflows/check.yml
jmao-denver: #160 is filed.

devinrsmith (approved):

PR #162, opened 2024-08-24
Find Zero-Assembler OpenJDK 21

AdrianBunk:
OpenJDK 21 on Debian uses `/usr/lib/jvm/default-java/lib/zero/` on Zero-Assembler architectures.


devinrsmith (approved):
Thanks for your contribution @AdrianBunk. I hadn't heard of https://openjdk.org/projects/zero/ before this. LGTM.

PR #164, opened 2024-08-30
fix: Make org.jpy.PyLib.getCurrentLocals/Globals work for Python 3.13

jmao-denver:
Python 3.13.0rc1 includes PEP667 (https://peps.python.org/pep-0667) and it made PyEval_GetLocals() no longer behave the same way as in Python 3.12 and older, which could be a bug to report to the cpython team but probably not worth it as it is marked as deprecated  together with PyEval_GetGlobals().

### DIFF
--- a/src/main/c/jni/org_jpy_PyLib.c
+++ b/src/main/c/jni/org_jpy_PyLib.c
@@ -559,7 +559,7 @@ JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_getCurrentGlobals
 
 #if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION <= 12
     globals = PyEval_GetGlobals(); // borrowed ref
-    JPy_INCREF(globals);
+    JPy_XINCREF(globals);
 #else
     // See https://peps.python.org/pep-0667 for the change in Python 3.13
     globals = PyEval_GetFrameGlobals(); // new ref
@@ -590,7 +590,7 @@ JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_getCurrentLocals
 
 #if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION <= 12
     locals = PyEval_GetLocals(); // borrowed ref
-    JPy_INCREF(locals);
+    JPy_XINCREF(locals);
 #else
     // See https://peps.python.org/pep-0667 for the change in Python 3.13
     locals = PyEval_GetFrameLocals(); // new ref
--- a/src/main/c/jni/org_jpy_PyLib.c
+++ b/src/main/c/jni/org_jpy_PyLib.c
@@ -521,6 +521,7 @@ PyObject *getMainGlobals() {
     }
 
     pyGlobals = PyModule_GetDict(pyMainModule); // borrowed ref
+    JPy_INCREF(pyGlobals);
 
     return pyGlobals;
 }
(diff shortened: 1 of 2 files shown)


devinrsmith (commented):
This seems like a good change. Should we prefer to instead _increment_ locals for python <= 3.12, and always XDECREF? At least from the PEPs example of migrating `PyEval_GetLocals`, it seems like incrementing the ref may be the more canonical approach?

```c
locals = PyEval_GetLocals();
if (locals == NULL) {
    goto error_handler;
}
Py_INCREF(locals);
```

to

```c
locals = PyEval_GetFrameLocals();
if (locals == NULL) {
    goto error_handler;
}
```

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -567,8 +571,11 @@ JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_getCurrentGlobals
     }
 
 error:
-    JPy_END_GIL_STATE
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION > 12
+    JPy_XDECREF(globals);
```
chipkent, line 575: I would add a comment about why this is needed for this version range to help us when we need to look back.

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -590,8 +602,11 @@ JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_getCurrentLocals
     }
 
 error:
-    JPy_END_GIL_STATE
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION > 12
+    JPy_XDECREF(locals);
```
chipkent, line 606: I would add a comment about why this is needed for this version range to help us when we need to look back.

jmao-denver:
> This seems like a good change. Should we prefer to instead _increment_ locals for python <= 3.12, and always XDECREF? At least from the PEPs example of migrating `PyEval_GetLocals`, it seems like incrementing the ref may be the more canonical approach?
> 
> ```c
> locals = PyEval_GetLocals();
> if (locals == NULL) {
>     goto error_handler;
> }
> Py_INCREF(locals);
> ```
> 
> to
> 
> ```c
> locals = PyEval_GetFrameLocals();
> if (locals == NULL) {
>     goto error_handler;
> }
> ```

I wasn't very comfortable with the existing pattern (use borrowed ref) on globals/locals either, but it wasn't critical when GIL guarantees the thread-safetly. Now with free-threaded Python is just around the corner it is time to use the correct pattern for borrowed refs.

chipkent (dismissed):

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -579,7 +588,14 @@ JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_getCurrentLocals
 
     JPy_BEGIN_GIL_STATE
 
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION <= 12
     locals = PyEval_GetLocals(); // borrowed ref
+    JPy_INCREF(locals);
```
devinrsmith, line 593: I think we should want this to be `JPy_XINCREF`? I see we don't have a macro for it right now, but should be easy to add.

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -555,7 +557,13 @@ JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_getCurrentGlobals
 
     JPy_BEGIN_GIL_STATE
 
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION <= 12
     globals = PyEval_GetGlobals(); // borrowed ref
+    JPy_INCREF(globals);
```
devinrsmith, line 562: ditto

devinrsmith (approved):

PR #165, opened 2024-09-17
feat: free-threaded Python (3.13.0+) support

### LINKED ISSUE #163: ENH: Support free-threaded python build
Add support for Python 3.13 nogil builds 

The experimental/preliminary work for numpy/Pandas/arrow seems a success:
https://github.com/numpy/numpy/issues/26157
https://github.com/pandas-dev/pandas/issues/59057
https://github.com/apache/arrow/issues/43536

jmao-denver:
Fixes #163


src/main/c/jpy_jobj.c
```diff
@@ -72,19 +72,30 @@ PyObject* JObj_FromType(JNIEnv* jenv, JPy_JType* type, jobject objectRef)
...
+    }
+#else
+    callable = PyDict_GetItemString(JPy_Type_Translations, type->javaName); // borrowed reference
+    JPy_XINCREF(callable);
+#endif
+
     if (callable != NULL) {
         if (PyCallable_Check(callable)) {
             callableResult = PyObject_CallFunction(callable, "OO", type, obj);
+            JPy_XDECREF(callable);
```
devinrsmith, line 90: This is maybe a question not directly related to this new feature; I don't think we use JPy_Type_Translations, but in the case where we _do_, should obj be DECREFd?

src/main/c/jpy_jtype.c
```diff
@@ -27,6 +27,54 @@
...
 #include "jpy_compat.h"
 
+#ifdef Py_GIL_DISABLED
+typedef struct {
+    PyMutex lock;
+    PyThreadState* owner;
+    int recursion_level;
+} ReentrantLock;
+
+static void acquire_lock(ReentrantLock* self) {
```
devinrsmith, line 37: Is this a pattern that you found suggested by C python, or elsewhere?

devinrsmith, line 37: `self` is an "interesting" choice; do we think `lock` would be better?

src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -153,45 +153,53 @@ def cleanup(references):
...
-                try {
-                    Thread.sleep(CLEANUP_THREAD_PASSIVE_SLEEP_MILLIS);
-                } catch (InterruptedException e) {
-                    Thread.currentThread().interrupt();
-                    return;
-                }
+            }
+        }
+        catch (RuntimeException e) {
+            if (!e.getMessage().contains("PyLib not initialized")) {
```
devinrsmith, line 201: `e.getMessage()` may be null.

src/main/java/org/jpy/PyObject.java
```diff
@@ -71,7 +71,7 @@ public static int cleanup() {
     PyObject(long pointer, boolean fromJNI) {
         state = new PyObjectState(pointer);
         if (fromJNI) {
-            if (CLEANUP_ON_INIT && PyLib.hasGil()) {
```
devinrsmith, line 74: I think the hasGil check is important... esp on older versions?

jpyutil.py
devinrsmith, line 1: I did not review this file yet, but I will next pass.

src/main/c/jpy_jtype.c
```diff
@@ -1015,19 +1082,29 @@ jboolean JType_AcceptMethod(JPy_JType* declaringClass, JPy_JMethod* method)
     PyObject* callableResult;
 
     //printf("JType_AcceptMethod: javaName='%s'\n", overloadedMethod->declaringClass->javaName);
+#ifdef Py_GIL_DISABLED
+    // if return is 1, callable is new reference
+    if (PyDict_GetItemStringRef(JPy_Type_Callbacks, declaringClass->javaName, &callable) != 1) {
+        callable = NULL;
+    }
+#else
+    callable = PyDict_GetItemString(JPy_Type_Callbacks, declaringClass->javaName); // borrowed reference
```
chipkent, line 1091:
```suggestion
    // borrowed ref, no need to replace with PyDict_GetItemStringRef because it is protected by the lock
    callable = PyDict_GetItemString(JPy_Type_Callbacks, declaringClass->javaName); 
```

src/main/c/jpy_jtype.c
```diff
@@ -151,6 +201,8 @@ JPy_JType* JType_GetType(JNIEnv* jenv, jclass classRef, jboolean resolve)
         return NULL;
     }
 
+    ACQUIRE_GET_TYPE_LOCK();
+    // borrowed ref, no need to replace with PyDict_GetItemRef because it protected by the lock
```
chipkent, line 205:
```suggestion
    // borrowed ref, no need to replace with PyDict_GetItemRef because it is protected by the lock
```

src/main/c/jpy_jtype.c
```diff
@@ -27,6 +27,54 @@
...
+    if (self->recursion_level == 0) {
+        self->owner = NULL;
+        PyMutex_Unlock(&(self->lock));
+    }
+}
+
+static ReentrantLock get_type_rlock = {{0}, NULL, 0};
+static ReentrantLock resolve_type_rlock = {{0}, NULL, 0};
+
+#define ACQUIRE_GET_TYPE_LOCK() acquire_lock(&get_type_rlock)
```
cpwright, line 67: TODO: Charles make sure we have no inversions.

src/main/c/jpy_jtype.c
```diff
@@ -1015,19 +1082,29 @@ jboolean JType_AcceptMethod(JPy_JType* declaringClass, JPy_JMethod* method)
     PyObject* callableResult;
 
     //printf("JType_AcceptMethod: javaName='%s'\n", overloadedMethod->declaringClass->javaName);
+#ifdef Py_GIL_DISABLED
+    // if return is 1, callable is new reference
+    if (PyDict_GetItemStringRef(JPy_Type_Callbacks, declaringClass->javaName, &callable) != 1) {
```
cpwright, line 1087: The GetItemStringRef is new in 3.13; but doesn't seem entirely related to the GIL-less changes.  It looks like a more robust API, should we use it always?

src/main/java/org/jpy/PyObject.java
```diff
@@ -71,7 +71,7 @@ public static int cleanup() {
...
+            if (CLEANUP_ON_INIT) {
                 REFERENCES.cleanupOnlyUseFromGIL(); // only performs *one* cleanup
```
cpwright, line 75: If we are no longer using the GIL, then we should rename this function and instead have comments on why it is correct.  If we are on 3.12 mode why are we still correct?

.github/workflows/build.yml
```diff
@@ -161,9 +160,105 @@ jobs:
...
+          path: dist/*.whl
+          retention-days: 1
+
+  bdist-wheels-linux-arm64-t:
+    runs-on: ${{ matrix.info.machine }}
+    strategy:
+      fail-fast: false
+      matrix:
+        info:
+          - { machine: 'ubuntu-20.04', python: '3.13t', arch: 'aarch64', cmd: '.github/env/Linux/bdist-wheel.sh' }
```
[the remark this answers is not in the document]
cpwright: 20.04 goes EOL 4/23/2025.  We may consider bumping to at least 22.04 which.

.github/workflows/build.yml
```diff
@@ -161,9 +160,105 @@ jobs:
...
+      matrix:
+        info:
+          - { machine: 'ubuntu-20.04', python: '3.13t', arch: 'amd64', cmd: '.github/env/Linux/bdist-wheel.sh' }
+          - { machine: 'macos-13', python: '3.13t', arch: 'amd64', cmd: '.github/env/macOS/bdist-wheel.sh' }
+          - { machine: 'macos-latest', python: '3.13t', arch: 'arm64', cmd: '.github/env/macOS/bdist-wheel.sh' }
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v3
```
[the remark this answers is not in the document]
jmao-denver: I agree with, it really depends on when GH actions add the free-threaded Python support.

.github/workflows/build.yml
```diff
@@ -79,7 +79,6 @@ jobs:
           distribution: 'temurin'
           java-version: '8'
 
-      - run: pip install setuptools
```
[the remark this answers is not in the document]
jmao-denver: Not relevant, just clean up a bit because the `cmd` script already makes sure all the needed packages are installed.

.github/workflows/build.yml
```diff
@@ -161,9 +160,105 @@ jobs:
...
 
+  bdist-wheels-t:
+    runs-on: ${{ matrix.info.machine }}
+    strategy:
+      fail-fast: false
+      matrix:
+        info:
+          - { machine: 'ubuntu-20.04', python: '3.13t', arch: 'amd64', cmd: '.github/env/Linux/bdist-wheel.sh' }
+          - { machine: 'macos-13', python: '3.13t', arch: 'amd64', cmd: '.github/env/macOS/bdist-wheel.sh' }
+          - { machine: 'macos-latest', python: '3.13t', arch: 'arm64', cmd: '.github/env/macOS/bdist-wheel.sh' }
```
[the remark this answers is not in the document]
jmao-denver: No universal2 wheel for FT build yet.

src/main/c/jpy_jtype.c
```diff
@@ -27,6 +27,54 @@
...
+        PyMutex_Unlock(&(self->lock));
+    }
+}
+
+static ReentrantLock get_type_rlock = {{0}, NULL, 0};
+static ReentrantLock resolve_type_rlock = {{0}, NULL, 0};
+
+#define ACQUIRE_GET_TYPE_LOCK() acquire_lock(&get_type_rlock)
+#define RELEASE_GET_TYPE_LOCK() release_lock(&get_type_rlock)
+#define ACQUIRE_RESOLVE_TYPE_LOCK() acquire_lock(&resolve_type_rlock)
```
cpwright, line 69: I think you need to add comments describing what these locks protect.

I think that you might be missing a case where the type dictionary is accessed in JType_AddField which has the resolve lock, but not the type lock.

Combining these two locks might be safer, as I think inversions could be possible as well.

.github/workflows/build.yml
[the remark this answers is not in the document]
jmao-denver: @devinrsmith is almost certainly changing to 22.04 will make the wheels not installation due to some incompatiblity in the glibc etc. Will file a follow-up ticket.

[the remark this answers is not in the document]
devinrsmith: We are currently targeting manylinux_2_17, which makes it compatible with GLIBC version 2.17+; afaict, the building of these artifacts needs to happen on a system w/ GLIBC version 2.17, and updating to 22.04 or later breaks this. We should have a ticket about GLIBC version support, and when we will move on past 2.17.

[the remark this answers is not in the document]
devinrsmith: https://github.com/pypa/cibuildwheel/issues/1772 has relevant upstream comments about the topic.

.github/workflows/build.yml
```diff
@@ -161,9 +160,105 @@ jobs:
...
+          - { machine: 'ubuntu-20.04', python: '3.13t', arch: 'aarch64', cmd: '.github/env/Linux/bdist-wheel.sh' }
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.21.3
```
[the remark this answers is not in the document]
jmao-denver: This still uses emulation but the nice thing about CBW under aarch64 simulation is that it can build both amd64 and aarch64 wheels.

.github/workflows/build.yml
```diff
@@ -161,9 +160,105 @@ jobs:
...
+          path: dist/*.whl
+          retention-days: 1
+
+  bdist-wheels-windows-t:
+    runs-on: ${{ matrix.info.machine }}
+    strategy:
+      fail-fast: false
+      matrix:
+        info:
+          - { machine: 'windows-2022', python: '3.13t', arch: 'amd64', cmd: '.\.github\env\Windows\bdist-wheel.ps1' }
```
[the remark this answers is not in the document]
jmao-denver: Yes, I am watching https://github.com/actions/setup-python/issues/771

src/main/c/jpy_jobj.c
jmao-denver: I think you are right.

src/main/c/jpy_jtype.c
jmao-denver: C Python doesn't offer an reentrant lock primitive. This is a pattern I found somewhere else.

src/test/python/jpy_mt_eval_exec_test.py
```diff
@@ -0,0 +1,121 @@
+import math
+import unittest
+
+import jpyutil
+
+jpyutil.init_jvm(jvm_maxmem='512M', jvm_classpath=['target/classes', 'target/test-classes'])
+import jpy
+# jpy.diag.flags = jpy.diag.F_TYPE
+
+NUM_THREADS = 20
```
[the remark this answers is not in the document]
jmao-denver: The standard GH hosted runner comes with 4 or fewer cores, https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories. 20 is really just a gut-feeling threshold but should be big enough to create contention among threads without incurring too much context switch. If we were to upgrade to larger runners (need to pay), we most likely need to make it a lot bigger to detect potential race conditions.

src/main/java/org/jpy/PyObject.java
jmao-denver: This change is made to fix a bug based on the two following observations:
1.  Because we now release the GIL whenever we cross into Java from Python (by @niloc132), the PyLib.hasGIL()  test will always fail in Python GIL enabled mode,  which results in that the cleanup method can never get called. (BTW, in the GIL disabled mode, PyLib.hasGIL() returns true).
2. As part of the same enhancement change by @niloc132, he made sure that the cleanup is done under GIL.
```
    private int cleanupOnlyUseFromGIL(long[] buffer) {
        return PyLib.ensureGil(() -> {
            int index = 0;
            while (index < buffer.length) {
                final Reference<? extends PyObject> reference = referenceQueue.poll();
                if (reference == null) {
   ...
```

niloc132: @jmao-denver is right, there are several layers of questionable here, and the latest round here is my fault. 
 * latest stable release: hasGil is always false here, so this block never executes
 * prior to the "release gil while in Java" change: hasGil is always _true_, so this _always_ runs.
 
Why did it even test to begin with? Why does it clean up all other object any time any PyObject is about to be passed in to Java (instead of just once either just before or just after calling Java)? Couldn't say. I think removing the check is correct (either just before the ft patch, or as part of it), and we should reevaluate what is even happening here separately.

Agreed on renaming cleanupOnlyUseFromGIL.

src/test/python/jpy_mt_eval_exec_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class MultiThreadedTestEvalExec(unittest.TestCase):
+    def setUp(self):
+        self.fixture = jpy.get_type("org.jpy.fixtures.MultiThreadedEvalTestFixture")
+        self.assertIsNotNone(self.fixture)
+
+    def atest_inc_baz(self):
+        baz = 15
+        self.fixture.script("baz = baz + 1; self.assertGreater(baz, 15)", NUM_THREADS)
```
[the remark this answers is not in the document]
jmao-denver: I think this is sufficient to ensure the multithreaded execution of a piece of code behaves the same when GIL is disabled and when it is enabled. Additional, more complex scripts in the tests only amount to testing that Python FT build is thread-safe just as the regular build.

src/main/c/jpy_module.c
```diff
@@ -646,23 +650,7 @@ PyObject* JPy_convert_internal(JNIEnv* jenv, PyObject* self, PyObject* args)
...
-    resultObj = (JPy_JObj*) PyObject_New(JPy_JObj, JTYPE_AS_PYTYPE(targetTypeParsed));
-    if (resultObj == NULL) {
-        (*jenv)->DeleteGlobalRef(jenv, objectRef);
-        return NULL;
-    }
-    // Store the reference to the converted object in the result JObj
-    ((JPy_JObj*) resultObj)->objectRef = objectRef;
-
-    return (PyObject*) resultObj;
+    return JObj_FromType(jenv, targetTypeParsed, objectRef);
```
jmao-denver, line 653: To fix a bug that failed to initialize `JPy_JArray` properly, which leads to a memory issue in the FT build, possibly due to some specific change in the Python C API (`PyType_GenericAlloc`) or the underlying memory management module used that no longer implicitly zero initialize the allocated memory.

I have tried to reproduce the problem with the GIL enabled build of <= 3.13 versions and confirmed that `JPy_JArray` has always be 'implicitly' zero initialized.

src/main/c/jpy_module.c
```diff
@@ -628,6 +628,7 @@ PyObject* JPy_convert_internal(JNIEnv* jenv, PyObject* self, PyObject* args)
         if (targetTypeParsed == NULL) {
             return NULL;
         }
+        JPy_DECREF(targetTypeParsed);
```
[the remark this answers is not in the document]
jmao-denver: This is a bit anti-pattern. But with the insider knowledge that the type is stored in a dictionary and guarded by the locks, it is safe to use. Will change to avoid the confusing.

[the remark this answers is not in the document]
chipkent: If it requires insider knowledge, it should have a comment at minimum.

.github/workflows/build.yml
```diff
@@ -157,13 +156,97 @@ jobs:
...
+  bdist-wheel-t:
+    runs-on: ${{ matrix.info.machine }}
+    strategy:
+      fail-fast: false
+      matrix:
+        info:
+          - { machine: 'ubuntu-20.04', python: '3.13t', java: '8', arch: 'amd64', cmd: '.github/env/Linux/bdist-wheel.sh' }
+          - { machine: 'windows-2022', python: '3.13t', java: '8', arch: 'amd64', cmd: '.\.github\env\Windows\bdist-wheel.ps1' }
+          - { machine: 'macos-13', python: '3.13t', java: '8', arch: 'amd64', cmd: '.github/env/macOS/bdist-wheel.sh' }
+          - { machine: 'macos-latest', python: '3.13t', java: '11', arch: 'arm64', cmd: '.github/env/macOS/bdist-wheel.sh' }
```
devinrsmith, line 172: I'm not a fan of using "latest"; let's pin this to `macos-14` for now, until we have reason to change it.

.github/workflows/check.yml
```diff
@@ -29,3 +29,32 @@ jobs:
...
+            java-version: ${{ matrix.java }}
+
+        - uses: astral-sh/setup-uv@v3
+        - run: |
+            uv python install ${{ matrix.python }}
+            uv venv --python ${{ matrix.python }}
+            source .venv/bin/activate
+            uv pip install pip
+            echo $JAVA_HOME
+            echo PATH=$PATH >> $GITHUB_ENV
```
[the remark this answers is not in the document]
devinrsmith: This looks like it is still using GITHUB_ENV?

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -499,12 +499,20 @@ void dumpDict(const char* dictName, PyObject* dict)
 
     size = PyDict_Size(dict);
     printf(">> dumpDict: %s.size = %ld\n", dictName, size);
+#ifdef Py_GIL_DISABLED
```
devinrsmith, line 502: I think the preference is to still use the critical section macros for 3.13+, even if the GIL is enabled. If the GIL is enabled, the macros simply expand to `{` / `}`.

If we want, we can define our own macro, `JPy_BEGIN_CRITICAL_SECTION` / `JPy_END_CRITICAL_SECTION` if we find we are need to use the pattern a lot.

jpyutil.py
```diff
@@ -326,18 +330,32 @@ def _find_python_dll_file(fail=False):
 
     # Prepare list of possible library file names
 
+    # account for Python debug builds
+    try:
+        sys.gettotalrefcount()
```
devinrsmith, line 335: I think

```python
import sysconfig

sysconfig.get_config_var('Py_DEBUG')
```

would be better?

It looks like technically debug ref counting is more specific than plain debug.

https://pythonextensionpatterns.readthedocs.io/en/latest/debugging/debug_python.html#the-debug-builds
https://pythonextensionpatterns.readthedocs.io/en/latest/debugging/debug_python.html#identifying-the-python-build-configuration-from-the-runtime

src/test/java/org/jpy/fixtures/MultiThreadedEvalTestFixture.java
```diff
@@ -0,0 +1,53 @@
...
+            }));
+        }
+        for (Thread thread : threads) {
+            thread.start();
+        }
+        for (Thread thread : threads) {
+            try {
+                thread.join();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
devinrsmith: If we are not going to throw an exception, probably best to explicitly return as well.

src/main/java/org/jpy/PyObject.java
devinrsmith: This was an oversight on my part during the "drop GIL" review.

I made the argument during the review that we might prefer to have some way to keep the GIL during "quick" java calls, or in other cases where we explicitly want to keep the GIL. I might argue that our call from `JType_CreateJavaPyObject` into the constructor of `PyObject` should keep the GIL.

The reason we need this to only be called into the context of the GIL (historically) is because we need the GIL when calling `PyLib.decRef`; I'm assuming in the case of free-threaded (/GIL-less) builds, Py_DECREF is thread safe. And it's not only that Py_DECREF that needs to be thread-safe, the arbitrary python deallocation code that can be executed as part of a decRef to 0 needs to be thread-safe.

In the context of free-threaded / GIL-less builds, maybe the functions need to be renamed to be something like `PyLib.isThreadSafe` / `PyLib.ensureThreadSafe`, or something like that.

chipkent (approved):

jpyutil.py
```diff
@@ -331,11 +331,8 @@ def _find_python_dll_file(fail=False):
     # Prepare list of possible library file names
 
     # account for Python debug builds
-    try:
-        sys.gettotalrefcount()
-        debug_build = True
-    except AttributeError:
-        debug_build = False
+
+    debug_build = sysconfig.get_config_var('Py_DEBUG')
```
devinrsmith, line 335: This is not correct in the context of how you are using it. It needs to be checked if it's set to "1"; at least in my builds, it's explicitly set to "0":

```
python3.8 -m sysconfig | grep Py_DEBUG    
        Py_DEBUG = "0"
```

After some more reading, I actually think that `sys.abiflags` is the more appropriate place to check, since it's _specifically_ mentioned as being set for debug builds, https://docs.python.org/3.10/using/configure.html#python-debug-build:

> Effects of a debug build:
> * Add `d` to [sys.abiflags](https://docs.python.org/3.10/library/sys.html#sys.abiflags).

src/main/java/org/jpy/PyObject.java
```diff
@@ -71,8 +71,8 @@ public static int cleanup() {
     PyObject(long pointer, boolean fromJNI) {
         state = new PyObjectState(pointer);
         if (fromJNI) {
-            if (CLEANUP_ON_INIT && PyLib.hasGil()) {
-                REFERENCES.cleanupOnlyUseFromGIL(); // only performs *one* cleanup
+            if (CLEANUP_ON_INIT) {
```
devinrsmith, line 74: Although originally unintentional (I think), this _is_ a change in behavior vs the current release. I might suggest Colin or I follow up with a PR, unless this is newly tested in this PR?

devinrsmith (changes requested):

jpyutil.py
devinrsmith: Sorry for my flip-flopping; it actually looks like `.get_config_var` is correct, since it seems to return an int when used like this. That said, maybe we should still prefer `sys.abiflags` since it's more explicitly documented?

jmao-denver: image

`Py_DEBUG` is also explicitly mentioned there. Also a quick ChatGPT inquiry yields this;

image

I think we should be comfortable with using  the value of `Py_DEBUG` in `sysconfig`

src/main/java/org/jpy/PyObject.java
jmao-denver: It isn't explicitly tested. But the code path does get executed implicitly in `jpy_typeconv_test_pyobj.py`.

jpyutil.py
devinrsmith: I'm not confident that `Py_DEBUG and Py_REF_DEBUG macros` are guaranteed to be equivalent to "config vars"; that said, it looks like `sys.abiflags` doesn't work on Windows, so I guess get_config_var is the best way for now.

It looks like there was a patch many years ago to add an official way to look up if it was built with debug, but it was not merged. https://github.com/python/cpython/issues/69443

devinrsmith (approved):

PR #172, opened 2024-10-08
Update Linux runners to ubuntu-24.04

devinrsmith:


devinrsmith:
Note, this is not possible as-is, we get the error:

> auditwheel: error: cannot repair "dist.linux/jpy-0.19.0.dev0-cp312-cp312-linux_x86_64.whl" to "manylinux_2_17_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.

PR #199, opened 2025-09-29
feat: 198: Call PyObject_IsTrue() to test truthiness

jmao-denver:


cpwright (approved):

src/main/c/jpy_conv.h
```diff
@@ -23,7 +23,7 @@
 extern "C" {
 #endif
 
-#define JPy_AS_JBOOLEAN(pyArg)   (jboolean) (pyArg == Py_True ? 1 : (pyArg == Py_False || pyArg == Py_None) ? 0 : (JPy_AS_CLONG(pyArg)) != 0)
+#define JPy_AS_JBOOLEAN(pyArg)   (jboolean)(PyObject_IsTrue(pyArg))
```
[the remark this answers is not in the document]
cpwright: I had similar thoughts.

src/test/python/jpy_array_test.py
```diff
@@ -58,6 +58,15 @@ def do_test_array_protocol_float2(self, type, initial, expected, places):
     def test_array_boolean(self):
         self.do_test_array_protocol('boolean', [False, False, False], [True, False, True])
 
+        # This seems a bit weird but numpy.ndarray allows this too, also but anyone familiar with strong typed languages
+        # shouldn't expect that values of types other than boolean can be assigned to a boolean array element.
+        jba = jpy.array("boolean", 3)
+        jba[0] = []
+        jba[1] = [1]
+        jba[2] = False
```
devinrsmith, line 66: Do we have numpy installed here? If so, would be good to test

```python
jba[3] = np.bool_(True)
jba[4] = np.bool_(False)
```

src/test/python/jpy_typeconv_test.py
```diff
@@ -318,6 +318,31 @@ def test_ToObjectArrayConversion(self):
...
+        self.assertTrue(jobj.getClass().equals(expected_type.jclass), f'Type is {jobj.getClass()}')
+        self.assertEqual(jpy.cast(jobj, expected_type).booleanValue(), False)
+
+        try:
+            import numpy as np
+        except ImportError:
+            return
+
+        ba = np.array([True, False])
+        jobj = jpy.convert(ba[0], java_boolean_type)
```
devinrsmith, line 341: I think `np.bool_(True)` is "better" than washing through array?

src/test/python/jpy_typeconv_test.py
```diff
@@ -318,6 +318,31 @@ def test_ToObjectArrayConversion(self):
...
+        self.assertEqual(jpy.cast(jobj, expected_type).booleanValue(), True)
+        jobj = jpy.convert(ba[1], java_boolean_type)
```
devinrsmith, line 343: I think `np.bool_(False)` is "better" than washing through array?

src/main/c/jpy_jtype.c
```diff
@@ -499,9 +499,8 @@ int JType_CreateJavaObject_2(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jcl
 int JType_CreateJavaBooleanObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jobject* objectRef)
 {
     jboolean value;
-    if (PyBool_Check(pyArg) || JPy_IS_CLONG(pyArg)) {
-        value = JPy_AS_JBOOLEAN(pyArg);
-    } else {
+    value = JPy_AS_JBOOLEAN(pyArg);
+    if ((signed char)value == -1) {
```
devinrsmith, line 503: This had a stricter type check, only working w/ bool or clong; now, we are explicitly making it work with anything that implements truthiness API... is this desirable in the context of `JType_CreateJavaBooleanObject`? (I don't know.)

src/main/c/jpy_conv.h
```diff
@@ -23,7 +23,7 @@
 extern "C" {
 #endif
 
-#define JPy_AS_JBOOLEAN(pyArg)   (jboolean) (pyArg == Py_True ? 1 : (pyArg == Py_False || pyArg == Py_None) ? 0 : (JPy_AS_CLONG(pyArg)) != 0)
+#define JPy_AS_JBOOLEAN(pyArg)   (jboolean) (pyArg == Py_True ? 1 : (pyArg == Py_False || pyArg == Py_None) ? 0 : PyObject_IsTrue(pyArg))
```
devinrsmith, line 26: What's the value in explicitly checking against Py_True, Py_False, and Py_None? Doesn't `PyObject_IsTrue` already handle this?

jmao-denver: aiming for a little performance gain, considering the majority of use cases

src/test/python/jpy_typeconv_test.py
jmao-denver: same as above.

PR #201, opened 2025-09-30
Remove EOL ubuntu-20.04 runners

devinrsmith:
The `ubuntu-20.04` runner images are no longer provided by GitHub actions, https://github.com/DMPRoadmap/roadmap/issues/3484. This means that there is no easy, native way to produce `manylinux_2_17` wheels on the native runner jobs.

To continue producing `manylinux_2_17` compatible wheels, this PR migrates the Linux wheel building process to the docker bake workflow which was already in use for the arm64 linux wheels. The  `bdist-wheel-linux-arm64` job was updated to an arm-native runner, which saw a ~50x speedup in job time.

In addition, the Windows / macOS free-threaded builds were rolled back into the normal `bdist-wheel` job since `actions/setup-python` now works with python version `3.13t`; this allows us to delete the special-case `bdist-wheel-t` and `bdist-wheel-linux-arm64-t` jobs.

A compromise in the name of build simplification was made to use the `ubuntu-22.04` / `ubuntu-22.04-arm` for the Linux `3.13t` wheels, which results in `manylinux_2_34` compatibility; it is not currently possible to use our docker bake workflow for these wheels.

The python 3.6, 3.7, and 3.8 wheels were removed from the build process given their EOL status, although, there was no fundamental change to jpy that necessitated this. (We may want to look at formally bumping the minimum python version in the future.)

### DIFF
--- a/.github/docker/docker-bake.hcl
+++ b/.github/docker/docker-bake.hcl
@@ -12,7 +12,7 @@ variable "DEBIAN_BASE" {
     # bullseye: manylinux2014 / manylinux_2_17
     # bookworm: manylinux_2_34
     # trixie: Unable to build on Debian trixie, https://github.com/jpy-consortium/jpy/issues/202
-    default = "trixie"
+    default = "bullseye"
 }
 
 variable "GITHUB_ACTIONS" {


.github/env/Linux/bdist-wheel.sh
```diff
@@ -14,8 +14,6 @@ python setup.py bdist_wheel --dist-dir dist.linux
 # Note: auditwheel only works with a single file argument - we are relying on finding exactly one wheel
 auditwheel \
   repair \
-  --plat "manylinux_2_17_$(arch)" \
-  --only-plat \
```
devinrsmith, line 18: Newer auditwheel versions can better infer the best minimum version based on the currently running OS, so best to let it decide.

jmao-denver (approved):
Tried installing the wheels on Mac and Windows, no issues found

PR #218, opened 2026-01-09
Add isGILEnabled Java method

### LINKED ISSUE #217: jpy should provide a Java method to check if Python GIL is enabled .

jmao-denver:
fixes #217

### DIFF
--- a/src/main/c/jni/org_jpy_PyLib.c
+++ b/src/main/c/jni/org_jpy_PyLib.c
@@ -173,6 +173,41 @@ JNIEXPORT jboolean JNICALL Java_org_jpy_PyLib_isGILEnabled
        // The sys._is_gil_enabled() function is the documented, stable way to check GIL status from both
        // Python code and C extensions in Python 3.13+ free-threaded builds.
 
+       jboolean result = JNI_TRUE;  // Default to GIL enabled
+       PyObject* sys_module = NULL;
+       PyObject* is_gil_enabled_func = NULL;
+       PyObject* call_result = NULL;
+
+       JPy_BEGIN_GIL_STATE(JNI_TRUE)
+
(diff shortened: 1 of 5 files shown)


cpwright (approved):

src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -149,6 +149,71 @@ JNIEXPORT jboolean JNICALL Java_org_jpy_PyLib_isPythonRunning
...
+    #ifdef Py_GIL_DISABLED
+       // Py_GIL_DISABLED indicates this is a free-threaded build that supports running without the GIL.
+       // We need to acquire the GIL before calling Python C API functions, even when checking if the GIL
+       // is enabled. This is because the Python C API is not thread-safe without the GIL.
+       //
+       // We call Python's sys._is_gil_enabled() instead of attempting to use internal C API functions.
+       // While Python 3.13+ has internal functions for checking GIL status, they are not part of the
+       // stable/public C API and might require including internal headers (e.g., pycore_*.h files).
+       // The sys._is_gil_enabled() function is the documented, stable way to check GIL status from both
+       // Python code and C extensions in Python 3.13+ free-threaded builds.
```
devinrsmith, line 174: I don't think this is _quite_ true. The documentation at https://docs.python.org/3/library/sys.html#sys._is_gil_enabled says:

> CPython implementation detail: It is not guaranteed to exist in all implementations of Python.

In general, I think the `_` prefix in a context like this means it's not part of the stable API.

I do think we should _strive_ to work with other implementations of Python.

That said, as you mention, it's possible that python doesn't provide a stable way to call this from C API yet.

I think we should at least note we'd prefer to call stable C API here, and create a follow-up in the future (/ link to our issue / link to c python issue).

PR #220, opened 2026-04-13
fix: Replace hand-rolled Python exception handling with  the Python built-in one.

### LINKED ISSUE #219: jpy only captures the top level Python exception which isn't very useful on chained exceptions.
Also the Python exception handling is complicated and brittle, we should utilize  the built-in  `traceback` module to be in line with Python's standard exception reporting and be future-proof.

jmao-denver:
fixes: #219

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,6 +86,7 @@ python_java_jpy_tests = [
     os.path.join(src_test_py_dir, 'jpy_field_test.py'),
     os.path.join(src_test_py_dir, 'jpy_retval_test.py'),
     os.path.join(src_test_py_dir, 'jpy_exception_test.py'),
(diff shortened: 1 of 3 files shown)


cpwright (approved):

devinrsmith (commented):
I don't have much to add; please see the copilot review to see if they are actually relevant or not (the NUL one may be a bit pedantic?).

Seems like delegating to traceback is a good idea.

jmao-denver:
> I don't have much to add; please see the copilot review to see if they are actually relevant or not (the NUL one may be a bit pedantic?).
> 
> Seems like delegating to traceback is a good idea.

Yeah, it isn't a new problem introduced by the change, and  a user almost has to deliberately make it happen.

PR #226, opened 2026-05-01
fix: Fix a race caused by premature-finalization when calling a method on transient PyObjects 

jmao-denver:


src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -213,7 +234,11 @@ public PyObject unwrap() {
      * @return the pointer to the underlying Python object wrapped by this dictionary.
      */
     long getPointer() {
-        return pyObject.getPointer();
+        try {
```
cpwright: I'm not sure why this actually needs a fence; because we are entirely within the context of the JVM.  The user of this method passing to PyLib would certainly need one though.

src/test/python/jpy_reachability_fence_test.py
```diff
@@ -0,0 +1,243 @@
...
+        try:
+            stress_fn()
+        finally:
+            self.Fixture.stopAllocator()
+
+    # ------------------------------------------------------------------
+    # Test: getAttribute + callMethod  (the original Deephaven crash path)
+    # ------------------------------------------------------------------
+    def test_stress_call_method(self):
+        """getAttribute('__call__') → transient.callMethod('__call__')"""
```
cpwright: I like to avoid the non-ASCII characters in our code.

src/test/java/org/jpy/fixtures/ReachabilityFenceTestFixture.java
```diff
@@ -0,0 +1,219 @@
...
+     */
+    public static void stressTypeChecks(PyObject obj, int iterations) {
+        for (int i = 0; i < iterations; i++) {
+            PyObject attr = obj.getAttribute("value");
+            attr.isInt();
+            attr.isFloat();
+            attr.isString();
+            attr.isCallable();
+            attr.isNone();
+            attr.isList();
```
cpwright: It think this sequence of usages actually means we are only testing the isDict call, which might be fine since the code is all the same; but we aren't doing what we intend I expect.

src/main/java/org/jpy/PyObject.java
```diff
@@ -441,8 +561,12 @@ public PyObject callMethod(String name, Object... args) {
     public PyObject call(String name, Object... args) {
         assertPythonRuns();
         Objects.requireNonNull(name, "name must not be null");
-        long pointer = PyLib.callAndReturnObject(getPointer(), false, name, args.length, args, null);
-        return pointer != 0 ? new PyObject(pointer) : null;
+        try {
+            long pointer = PyLib.callAndReturnObject(getPointer(), false, name, args.length, args, null);
+            return pointer != 0 ? new PyObject(pointer) : null;
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith, line 570: I'm going to be a bit pedantic, since we are using some low-level methods here. We should only be ensuring reachability for the lifetime of the JNI call. 

```java
        final long pointer;
        try {
            pointer = PyLib.callAndReturnObject(getPointer(), false, name, args.length, args, null);
        } finally {
            Reference.reachabilityFence(this);
        }
       return pointer != 0 ? new PyObject(pointer) : null;
```

src/main/java/org/jpy/PyObject.java
```diff
@@ -424,8 +540,12 @@ public <T> void setAttribute(String name, T value, Class<? extends T> valueType)
     public PyObject callMethod(String name, Object... args) {
         assertPythonRuns();
         Objects.requireNonNull(name, "name must not be null");
-        long pointer = PyLib.callAndReturnObject(getPointer(), true, name, args.length, args, null);
-        return pointer != 0 ? new PyObject(pointer) : null;
+        try {
+            long pointer = PyLib.callAndReturnObject(getPointer(), true, name, args.length, args, null);
+            return pointer != 0 ? new PyObject(pointer) : null;
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith, line 548: This reachability fence can be tightened.

src/main/java/org/jpy/PyObject.java
```diff
@@ -332,8 +424,12 @@ public <T> T[] getObjectArrayValue(Class<? extends T> itemType) {
     public PyObject getAttribute(String name) {
         assertPythonRuns();
         Objects.requireNonNull(name, "name must not be null");
-        long pointer = PyLib.getAttributeObject(getPointer(), name);
-        return pointer != 0 ? new PyObject(pointer) : null;
+        try {
+            long pointer = PyLib.getAttributeObject(getPointer(), name);
+            return pointer != 0 ? new PyObject(pointer) : null;
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith, line 431: This reachability fence can be tightened.

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -222,6 +243,11 @@ long getPointer() {
      * @return a wrapped copy of this Python dictionary.
      */
     public PyDictWrapper copy() {
-        return new PyDictWrapper(PyLib.copyDict(pyObject.getPointer()));
+        try {
+            return new PyDictWrapper(PyLib.copyDict(pyObject.getPointer()));
+        } finally {
+            Reference.reachabilityFence(this);
+        }
```
devinrsmith, line 264: This reachability fence can be tightened - we might consider adding copyDict to PyObject (can be package private if we don't want to make public).

Should only need to ensure reachability for pyObject (not this, right? this may be more already accomplished if we add copyDict to PyObject).

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -195,6 +214,8 @@ public Set<Entry<PyObject, PyObject>> entrySet() {
                 .stream()
                 .map(p -> new SimpleImmutableEntry<>(p, get(p)))
                 .collect(Collectors.toCollection(LinkedHashSet::new));
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith: This reachability fence can be tightened - we might consider adding pyDictKeys to PyObject (can be package private if we don't want to make public).

Should only need to ensure reachability for pyObject (not this, right? this may be more already accomplished if we add pyDictKeys to PyObject).

cpwright (dismissed):

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -174,7 +189,11 @@ public Set<PyObject> keySet() {
      */
     @Override
     public Collection<PyObject> values() {
-        return PyLib.pyDictValues(this.pyObject.getPointer()).asList();
+        try {
+            return PyLib.pyDictValues(this.pyObject.getPointer()).asList();
+        } finally {
+            Reference.reachabilityFence(this);
+        }
```
devinrsmith, line 202: This reachability fence can be tightened - we might consider adding pyDictValues to PyObject (can be package private if we don't want to make public).

Should only need to ensure reachability for pyObject (not this, right? this may be more already accomplished if we add pyDictValues to PyObject).

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -164,6 +177,8 @@ public void clear() {
     public Set<PyObject> keySet() {
         try (final PyObject pyObj = PyLib.pyDictKeys(this.pyObject.getPointer())) {
             return new LinkedHashSet<>(pyObj.asList());
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith: This reachability fence can be tightened - we might consider adding pyDictKeys to PyObject (can be package private if we don't want to make public).

Should only need to ensure reachability for pyObject (not this, right? this may be more already accomplished if we add pyDictKeys to PyObject).

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -54,18 +55,30 @@ public boolean isEmpty() {
...
+        }
     }
 
     public boolean containsKey(PyObject key) {
-        return PyLib.pyDictContains(pyObject.getPointer(), key, PyObject.class);
+        try {
+            return PyLib.pyDictContains(pyObject.getPointer(), key, PyObject.class);
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith: Should only need to ensure reachability for pyObject, right?

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -54,18 +55,30 @@ public boolean isEmpty() {
...
 
     /**
      * An extension to the Map interface that allows the use of String keys without generating warnings.
      */
     public boolean containsKey(String key) {
-        return PyLib.pyDictContains(pyObject.getPointer(), key, String.class);
+        try {
+            return PyLib.pyDictContains(pyObject.getPointer(), key, String.class);
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith: Should only need to ensure reachability for pyObject, right?

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -54,18 +55,30 @@ public boolean isEmpty() {
 
     @Override
     public boolean containsKey(Object key) {
-        return PyLib.pyDictContains(pyObject.getPointer(), key, null);
+        try {
+            return PyLib.pyDictContains(pyObject.getPointer(), key, null);
+        } finally {
+            Reference.reachabilityFence(this);
```
devinrsmith: Should only need to ensure reachability for pyObject, right?

devinrsmith (commented):
All of my comments are about limiting the scope of reachability.

src/main/java/org/jpy/PyDictWrapper.java
jmao-denver: yeah, but feels like a force change for the sake of making RF code centralized.

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -164,6 +177,8 @@ public void clear() {
     public Set<PyObject> keySet() {
         try (final PyObject pyObj = PyLib.pyDictKeys(this.pyObject.getPointer())) {
             return new LinkedHashSet<>(pyObj.asList());
+        } finally {
+            Reference.reachabilityFence(this.pyObject);
```
devinrsmith, line 182: This is too broad - we should only need to have reachability for the duration of PyLib.pyDictKeys.

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -174,7 +189,11 @@ public Set<PyObject> keySet() {
      */
     @Override
     public Collection<PyObject> values() {
-        return PyLib.pyDictValues(this.pyObject.getPointer()).asList();
+        try {
+            return PyLib.pyDictValues(this.pyObject.getPointer()).asList();
+        } finally {
+            Reference.reachabilityFence(this.pyObject);
+        }
```
devinrsmith, line 202: This is too broad - we should only need to have reachability for the duration of PyLib.pyDictValues (the code here also encompasses asList which it shouldn't need to do).

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -195,6 +214,8 @@ public Set<Entry<PyObject, PyObject>> entrySet() {
                 .stream()
                 .map(p -> new SimpleImmutableEntry<>(p, get(p)))
                 .collect(Collectors.toCollection(LinkedHashSet::new));
+        } finally {
+            Reference.reachabilityFence(this.pyObject);
```
devinrsmith: This is too broad - we should only need to have reachability for the duration of PyLib.pyDictKeys.

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -222,6 +243,11 @@ long getPointer() {
      * @return a wrapped copy of this Python dictionary.
      */
     public PyDictWrapper copy() {
-        return new PyDictWrapper(PyLib.copyDict(pyObject.getPointer()));
+        try {
+            return new PyDictWrapper(PyLib.copyDict(pyObject.getPointer()));
+        } finally {
+            Reference.reachabilityFence(this.pyObject);
+        }
```
devinrsmith, line 264: This is too broad - we should only need to have reachability for the duration of PyLib.copyDict.

src/main/java/org/jpy/PyDictWrapper.java
devinrsmith: The reachability fence can be tightened just in this code too, w/o centralization.

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -222,6 +257,15 @@ long getPointer() {
...
      */
     public PyDictWrapper copy() {
-        return new PyDictWrapper(PyLib.copyDict(pyObject.getPointer()));
+        final PyObject pyObj;
+        try {
+            pyObj = PyLib.copyDict(this.pyObject.getPointer());
+        } finally {
+            Reference.reachabilityFence(this.pyObject);
+        }
+        try (pyObj) {
```
devinrsmith: This is a change from previous semantics - do we want to close pyObj?

src/main/java/org/jpy/PyDictWrapper.java
```diff
@@ -174,7 +194,15 @@ public Set<PyObject> keySet() {
...
     @Override
     public Collection<PyObject> values() {
-        return PyLib.pyDictValues(this.pyObject.getPointer()).asList();
+        final PyObject pyObj;
+        try {
+            pyObj = PyLib.pyDictValues(this.pyObject.getPointer());
+        } finally {
+            Reference.reachabilityFence(this.pyObject);
+        }
+        try (pyObj) {
```
devinrsmith: This is a change from previous semantics - do we want to close pyObj?

jmao-denver: new ArrayList makes a copy, right? If so, we can be proactive?

src/main/java/org/jpy/PyDictWrapper.java
jmao-denver: Yeah, should keep it alive.

cpwright (dismissed):

cpwright (approved):

devinrsmith (approved):
LGTM

PR #233, opened 2026-05-15
refactor:  simplify Python object cleanup to make it more efficient and avoid accessing Java heap when Java OOM happens

jmao-denver:


src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -74,33 +88,59 @@ private Reference<PyObject> asRef(PyObject pyObject) {
...
-     * This should *only* be invoked through the proxy, or when we *know* we have the GIL.
+     * Drains the reference queue and decrements the Python reference counts of all collected
+     * {@link PyObject} instances in a single batch.
+     *
+     * <p>The GIL is acquired internally by {@link PyLib#decRef} and {@link PyLib#decRefs}, so
+     * callers do not need to hold the GIL before invoking this method. The reference-queue
+     * draining is pure Java and requires no GIL.
+     *
+     * <p>This method performs zero Java heap allocation inside any JNI call, making it safe
+     * to invoke even under severe heap pressure.
```
devinrsmith: This is not true. Python destructors can call back into java and execute arbitrary code.

src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -119,85 +159,53 @@ private int appendIfNotClosed(long[] buffer, int index, Reference<? extends PyOb
...
-                // It would be much nicer if ReferenceQueue exposed a method that blocked until the
-                // queue was non-empty and *doesn't* remove any items. We can potentially implement this
-                // by using reflection to access the internal lock of the ReferenceQueue in the future.
-
-                if (size == buffer.length) {
-                    if (CLEANUP_THREAD_ACTIVE_SLEEP_MILLIS == 0) {
-                        Thread.yield();
+        return new Thread(() -> {
+            try {
+                boolean backlogged = false;
```
devinrsmith: Not a fan of using an anonymous lamba for this, I think makes it a little harder to trace stacktrace. Give this a proper method?

src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -119,85 +159,53 @@ private int appendIfNotClosed(long[] buffer, int index, Reference<? extends PyOb
...
+                while (!Thread.currentThread().isInterrupted()) {
+                    final Reference<? extends PyObject> first;
+                    if (backlogged) {
+                        // Previous batch was full — more items are likely queued.
+                        // Poll immediately rather than blocking, to drain quickly.
+                        first = referenceQueue.poll();
```
devinrsmith: There should be no difference in performance. I would stick to the cleaner code that is more streamlined; that is, always doing a blocking poll on the first call after a batch.

src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -74,33 +88,59 @@ private Reference<PyObject> asRef(PyObject pyObject) {
...
-                    break;
-                }
-                index = appendIfNotClosed(buffer, index, reference);
-            }
-            if (index == 0) {
-                return 0;
+        int index = 0;
+        while (index < buffer.length) {
+            final Reference<? extends PyObject> reference = referenceQueue.poll();
+            if (reference == null) {
```
devinrsmith: I would have this method delegate to the `threadSafeCleanup(first)`.

src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -119,85 +159,53 @@ private int appendIfNotClosed(long[] buffer, int index, Reference<? extends PyOb
...
                     } else {
-                        try {
-                            Thread.sleep(CLEANUP_THREAD_ACTIVE_SLEEP_MILLIS);
-                        } catch (InterruptedException e) {
-                            Thread.currentThread().interrupt();
-                            return;
+                        // Block until an item arrives or the timeout expires.
+                        // This avoids busy-polling and wakes up immediately when
+                        // there is cleanup work to do.
+                        first = referenceQueue.remove(CLEANUP_THREAD_PASSIVE_SLEEP_MILLIS);
```
devinrsmith: I'll note, the previous code had this comment:

```java
                // It would be much nicer if ReferenceQueue exposed a method that blocked until the
                // queue was non-empty and *doesn't* remove any items.
```

it makes me wonder why we didn't use `remove(timeout)` in the first place.

src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -74,33 +88,59 @@ private Reference<PyObject> asRef(PyObject pyObject) {
...
+            return 1;
+        }
+        PyLib.decRefs(buffer, index);
+        return index;
+    }
 
-            // We really really really want to make sure we *already* have the GIL lock at this point in
-            // time. Otherwise, we block here until the GIL is available for us, and stall all cleanup
-            // related to our PyObjects.
-
```
devinrsmith, line 96: I'm worried there is some reason this was _necessary_, but I can't remember why... b/c, calling PyLib.decRef _does_ take the GIL...

src/test/python/jpy_cleanup_thread_test.py
```diff
@@ -0,0 +1,351 @@
...
+import threading
+import time
+import unittest
+
+import jpyutil
+
+jpyutil.init_jvm(
+    jvm_maxmem='256M',
+    jvm_classpath=['target/classes', 'target/test-classes'],
+)
```
devinrsmith, line 43: Your tests explicitly reference G1 GC; should this test explicitly set G1 as garbage collector?

src/test/python/jpy_cleanup_thread_test.py
```diff
@@ -0,0 +1,351 @@
...
+        Reproduces the production scenario from hs_err_pid89198.log: the JVM
+        runs a mandatory full GC before throwing OutOfMemoryError, which enqueues
+        pending WeakReferences as a side effect.  The cleanup daemon wakes up via
+        ReferenceQueue.remove() and calls PyLib.decRefs() while OOM conditions exist.
+
+        Under the old code this crashed: the proxy round-trip caused
+        JPy_HandleJavaException to call Throwable.getStackTrace() on the OOM
+        exception, triggering a G1 read barrier on null → SIGSEGV.
+        The fix removes all Java heap allocation from the cleanup path —
+        PyLib.decRefs() is void-returning and allocation-free in JNI.
```
devinrsmith: I'm a bit confused exactly what happened in the old code... something is trying to deref a null pointer? That seems like a logical error and not something necessarily something that implicates the old code that was taking GIL up-front...

src/test/java/org/jpy/fixtures/CleanupThreadTestFixture.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Releases all held {@link PyObject} wrappers.  The caller is responsible
+     * for triggering GC (e.g. via {@code System.gc()}) if it needs the
+     * {@link java.lang.ref.WeakReference}s to be enqueued promptly.
+     */
+    public void releaseAll() {
+        held.clear();
+        held = new ArrayList<>(); // drop old list to remove any lingering reference
```
devinrsmith: It's a bit odd to do both of these things; either should accomplish the goal you want.

src/main/java/org/jpy/PyObjectReferences.java
devinrsmith: Even though ReferenceQueue is thread safe, it's a _little_ odd to access it without the lock. We may have been abusing the GIL in this way originally (to ensure exclusive access). And I see previously, there was a synchronized on threadSafeCleanup.

This may be the reason why we didn't use the timeout variant in the past? We don't want to block callers while we are sleeping...

src/main/java/org/jpy/PyObject.java
```diff
@@ -76,11 +106,10 @@ public static int cleanup() {
...
         state = new PyObjectState(pointer);
-        if (fromJNI) {
-            if (CLEANUP_ON_THREAD) {
-                // ensures that we've only started after python has been started, and we know there is something to cleanup
-                startCleanupThread();
-            }
+        // Start the cleanup daemon lazily on the first PyObject arriving from JNI;
+        // fromJNI==true guarantees Python is already running (the C layer sets JNI_TRUE).
+        if (fromJNI && CLEANUP_ON_THREAD && CLEANUP_THREAD.get() == null) {
+            startCleanupThread();
```
devinrsmith: You are double-checking `CLEANUP_THREAD.get() == null`; I think just call the method and let it check. (You can re-name the method `maybeStartCleanupThread`, or similar, to make it a little clearer if you want.)

src/main/java/org/jpy/PyLib.java
```diff
@@ -240,17 +240,29 @@ public static void startPython(int flags, String... extraPaths) {
...
+     * Environment originating from the Python interpreter ({@code Py_Finalize} in standard
+     * CPython). There is currently no workaround other than not restarting the interpreter.
+     * For more information refer to <a href="https://github.com/bcdev/jpy/issues/70">issue #70</a>.
      */
     public static void stopPython() {
-        PyObject.cleanup();
+        // Stop the daemon before Py_Finalize — it must not call decRef/decRefs against a
+        // dead interpreter. The final drain in stopCleanupThread() covers any references
+        // enqueued after the daemon's last poll.
+        PyObject.stopCleanupThread();
```
devinrsmith, line 265: This is a clear improvement. Sad I didn't think of this.

src/main/java/org/jpy/PyObject.java
```diff
@@ -64,7 +64,37 @@ private static void startCleanupThread() {
...
+                Thread.currentThread().interrupt();
+            }
+        }
+        // Final drain: pick up any WeakReferences enqueued between the daemon's last poll and
+        // being interrupted.  This is hygiene rather than crash prevention — Py_Finalize() will
+        // reclaim all Python memory regardless of Java-side refcounts, and unprocessed entries
+        // in the queue will simply be discarded at JVM shutdown without causing any late decRef
+        // calls.  The drain ensures reference counts are orderly at shutdown and avoids leaving
+        // stale entries in the REFERENCES map.
+        REFERENCES.cleanup();
```
devinrsmith: I don't think it matters technically in the current workflow, but I think this cleanup should only happen if `thread != null` (should not matter whether `thread.isAlive` or not).

setup.py
```diff
@@ -235,9 +245,7 @@ def _write_jpy_config(target_dir=None, install_dir=None):
     Write out a well-formed jpyconfig.properties file for easier Java
     integration in a given location.
     """
-    if is_ci:
-        # We don't want to publish the properties for the CI build system.
-        return None
```
devinrsmith, line 240: Why was this changed?

.github/workflows/check.yml
```diff
@@ -61,7 +61,7 @@ jobs:
             source .venv/bin/activate
             uv pip install pip
             echo $JAVA_HOME
-            echo PATH=$PATH >> $GITHUB_PATH
+            echo "$PWD/.venv/bin" >> $GITHUB_PATH
```
devinrsmith: Is sourcing the venv important now?

If we are passing the explicit bin to GITHUB_PATH now, should we just invoke:

```bash
uv pip install pip --python .venv/
```

src/main/java/org/jpy/PyObject.java
jmao-denver: Double checking only happens once though.

setup.py
jmao-denver: To run the now re-activated Java->Py tests, we need to generate the `jpyconfig.properties` file in CI as well.

.github/workflows/check.yml
jmao-denver: It was needed because the venv path wasn't added correctly. But we should really have used GH actions' native Python action for FT Python versions once they were supported.

src/main/java/org/jpy/PyObject.java
devinrsmith: Sure, but it's the same as writing:

```java
if (fromJNI && CLEANUP_ON_THREAD && CLEANUP_THREAD.get() == null && CLEANUP_THREAD.get() == null) {
  // ...
}
```
the double check only happens once, but it's extraneous either way.

setup.py
devinrsmith: Does this effect the released artifact though? I think so?

jmao-denver: No, the path is never taken when building the artifacts.

src/main/java/org/jpy/PyObject.java
jmao-denver: This is done to avoid the method call each time the constructor is called which could happen a lot. Tiny savings but maybe worth it?

setup.py
devinrsmith: I don't think that's true.

Looking at the logs now, I see we are hitting this path and it is erroring out. I don't know if this is new behavior or not...

```

jpyutil - INFO: jpy Python API configuration written to '/home/runner/work/jpy/jpy/build/lib.linux-x86_64-3.13/jpyconfig.py'
INFO: jpy Python API configuration written to '/home/runner/work/jpy/jpy/build/lib.linux-x86_64-3.13/jpyconfig.py'
jpyutil - ERROR: Error while writing Java API configuration
  File "/home/runner/work/jpy/jpy/build/lib.linux-x86_64-3.13/jpyutil.py", line 669, in write_config_files
    java_api_properties = _get_java_api_properties(fail=req_java_api_conf, path=install_dir)
  File "/home/runner/work/jpy/jpy/build/lib.linux-x86_64-3.13/jpyutil.py", line 125, in _get_java_api_properties
    jpy_config.set_property('jpy.jpyLib', _get_module_path('jpy', fail=fail, install_path=path))
                                          ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/jpy/jpy/build/lib.linux-x86_64-3.13/jpyutil.py", line 108, in _get_module_path
    raise ImportError("module '{}' not found, or is not loadable".format(name))
ImportError: module 'jpy' not found, or is not loadable
```

src/main/java/org/jpy/PyObject.java
devinrsmith: We should not be worried about calling a method.

devinrsmith (approved):

cpwright (approved):

PR #236, opened 2026-06-02
fix: fix the jpy_cleanup_thread_test for 3.14t

jmao-denver:
1. the fix to the test solves a 'recursion depth limit exceeded' error in the test itself due to the use of lock and its being used as a safe point to deallocate refcount=zero object in 3.14t
2. due to some unknown networking hiccups in GH Actions, the CI run regularly fails when downloading external deps. The added retry-logic and the removal of unused deps in pom.xml should make the run more stable.

### DIFF
--- a/.github/docker/Dockerfile
+++ b/.github/docker/Dockerfile
@@ -16,11 +16,12 @@ RUN \
     echo "deb [signed-by=/usr/share/keyrings/adoptium.asc] https://packages.adoptium.net/artifactory/deb $(awk -F= '/^VERSION_CODENAME/{print$2}' /etc/os-release) main" | tee /etc/apt/sources.list.d/adoptium.list; \
     apt-get -qq update; \
     for i in 1 2 3 4 5; do \
-        apt-get -qq -y --no-install-recommends --fix-missing install temurin-11-jdk maven && break; \
+        apt-get -qq -y --no-install-recommends install temurin-11-jdk maven && break; \
         echo "apt-get install attempt $i failed; retrying in 5s..."; \
         sleep 5; \
         apt-get -qq update || true; \
-    done
+    done; \
+    command -v mvn >/dev/null && command -v javac >/dev/null
 RUN \
     --mount=type=cache,target=/root/.cache/pip,sharing=locked \
     set -eux; \
--- a/.github/docker/Dockerfile
+++ b/.github/docker/Dockerfile
@@ -12,16 +12,15 @@ RUN \
     set -eux; \
     rm -f /etc/apt/apt.conf.d/docker-clean; \
     echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' | tee /etc/apt/apt.conf.d/keep-cache; \
-    wget --no-hsts -O /usr/share/keyrings/adoptium.asc https://packages.adoptium.net/artifactory/api/gpg/key/public; \
+    echo 'Acquire::Retries "3";'           >  /etc/apt/apt.conf.d/80-retries; \
+    echo 'Acquire::Retries::Delay "true";'  >> /etc/apt/apt.conf.d/80-retries; \
+    echo 'Acquire::http::Timeout "30";'     >> /etc/apt/apt.conf.d/80-retries; \
+    echo 'Acquire::https::Timeout "30";'    >> /etc/apt/apt.conf.d/80-retries; \
+    wget --no-hsts --tries=3 --waitretry=5 --retry-connrefused --timeout=10 \
+         -O /usr/share/keyrings/adoptium.asc https://packages.adoptium.net/artifactory/api/gpg/key/public; \
     echo "deb [signed-by=/usr/share/keyrings/adoptium.asc] https://packages.adoptium.net/artifactory/deb $(awk -F= '/^VERSION_CODENAME/{print$2}' /etc/os-release) main" | tee /etc/apt/sources.list.d/adoptium.list; \
     apt-get -qq update; \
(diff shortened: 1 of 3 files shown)


.github/docker/Dockerfile
```diff
@@ -15,7 +15,13 @@ RUN \
...
     echo "deb [signed-by=/usr/share/keyrings/adoptium.asc] https://packages.adoptium.net/artifactory/deb $(awk -F= '/^VERSION_CODENAME/{print$2}' /etc/os-release) main" | tee /etc/apt/sources.list.d/adoptium.list; \
     apt-get -qq update; \
-    apt-get -qq -y --no-install-recommends install temurin-11-jdk maven
+    for i in 1 2 3 4 5; do \
+        apt-get -qq -y --no-install-recommends install temurin-11-jdk maven && break; \
+        echo "apt-get install attempt $i failed; retrying in 5s..."; \
+        sleep 5; \
+        apt-get -qq update || true; \
+    done; \
+    command -v mvn >/dev/null && command -v javac >/dev/null
```
devinrsmith: This seems fragile, I don't know if it's something I would want to commit.

src/test/python/jpy_cleanup_thread_test.py
```diff
@@ -55,40 +55,49 @@
...
+#   * GIL Python: list.append is a single C call that holds the GIL for its
+#     full duration → atomic even with two concurrent writers (test 06).
+#     len(list) reads ob_size in one C call → atomic.
+#   * Freethreaded 3.14t: list.append uses the list's per-object critical
+#     section. It's only a safe point on contention; deferred-decref keeps
+#     __init__ and __del__ single-writer per counter, so the lock is
+#     uncontested → no safe point in __del__ → no recursion cascade.
+
+_created   = []
+_destroyed = []
```
devinrsmith: Interesting. I see this ticket - https://github.com/python/cpython/issues/124366, maybe worthwhile reference here as a possible improvement in the future?

pom.xml
```diff
@@ -116,7 +116,6 @@
                     <destDir>java-apidocs</destDir>
                     <doctitle>${project.name} ${project.version} Java API</doctitle>
                     <windowtitle>${project.name} ${project.version} Java API</windowtitle>
-                    <excludePackageNames>org.jpy.annotations</excludePackageNames>
```
devinrsmith, line 119: Why change this?

src/test/python/jpy_cleanup_thread_test.py
```diff
@@ -55,40 +55,49 @@
...
-_lock      = threading.RLock()
-_created   = 0
-_destroyed = 0
+# Counters must work on both GIL-enabled CPython and freethreaded CPython 3.14t.
+#
+# Why a lock is *not* used:
+#   On freethreaded 3.14t, acquiring any Python-level lock inside __del__ is a
+#   safe point. That lets the interpreter drain pending deferred decrefs, which
+#   fires more __del__ calls, which re-acquire the same lock — leading to
+#   RecursionError. The cleanup daemon avoids this by deferring Py_DECREF back
```
devinrsmith: Is there any python documentation on this?

pom.xml
jmao-denver: Line 124 already covers it.

.github/docker/Dockerfile
jmao-denver: It is probably as fragile as the existing code? I am fine with leaving it alone if we can live with the annoying transient networking issue and having to rerun the jobs manually.

cpwright: The alternative is we use a prebuilt docker image; with exactly the packages we want.  That is how core works, but it adds extra dependencies and jobs to the repo if we go that route.

The Deephaven Enterprise builds also have problems with transient failures on remote repositories like these.

src/test/python/jpy_cleanup_thread_test.py
```diff
@@ -55,40 +55,51 @@
...
+#     under the GIL, so it is atomic even when two threads append concurrently
+#     (see test_06, where the cleanup daemon and the main thread both fire
+#     __del__).  len(list) is also a single C call → atomic.
+#   * Free-threaded Python — list.append() takes the list's per-object critical
+#     section.  That lock is only a safe point on contention; the cleanup
+#     daemon defers Py_DECREF back to the owning thread, so each counter has a
+#     single writer in steady state and the critical section stays on its
+#     uncontested fast path.  Acquiring a Python-level lock (threading.Lock /
+#     RLock) inside __del__ would be a safe point and could drain queued
+#     decrefs recursively — see doc/freethreaded-safe-points.md.
```
cpwright: The doc/freethreaded-safe-points.md is not something you've included in the repo so we should not reference it from this comment.

.github/docker/Dockerfile
```diff
@@ -12,16 +12,15 @@ RUN \
     set -eux; \
     rm -f /etc/apt/apt.conf.d/docker-clean; \
     echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' | tee /etc/apt/apt.conf.d/keep-cache; \
-    wget --no-hsts -O /usr/share/keyrings/adoptium.asc https://packages.adoptium.net/artifactory/api/gpg/key/public; \
```
cpwright, line 15: why remove the public key thing?

jmao-denver: moved, not removed.

devinrsmith (approved):

cpwright (approved):

